PR #1, opened 2025-08-08
feat: csv to json workflow

dumikau:
This commit introduces a workflow that compiles data for each network and category into JSON files, saved as `json/<network_name>.json` on the `json` branch.

Source data is organized in CSV files under:
`networks/<network_name>/<category_name>.csv`

Each CSV may include a provider column with references like: `!provider:<provider_slug>`, which points to shared data in `providers/<category_name>.csv`.

When a provider reference is present, any empty fields in the network CSV are automatically overridden with values from the referenced provider.


networks/flare/rpc.csv
```diff
@@ -0,0 +1,16 @@
...
+ankr-mainnet-free,!provider:ankr-free-recent-state,,,,,,,,,,,,,,,,,,,
+ankr-mainnet-growth,!provider:ankr-pay-as-you-go-recent-state ,,,,,,,,,,,,,,,,,,,
+ankr-mainnet-enterprise,!provider:ankr-enterprise-recent-state,,,,,,,,,,,,,,,,,,,
+quicknode-coston2-free,!provider:quicknode-free-recent-state,,,,,,,,,,,,,,,,,,,
+quicknode-mainnet-free,!provider:quicknode-free-recent-state,,,,,,,,,,,,,,,,,,,
+quicknode-mainnet-growth,!provider:quicknode-pay-as-you-go-recent-state,,,,,,,,,,,,,,,,,,,
+quicknode-mainnet-enterprise,!provider:quicknode-enterprise-recent-state,,,,,,,,,,,,,,,,,,,
+nownodes-mainnet-free,NOWNodes,Free,Recent-State,mainnet,0,"Free (100,000 included only)",null,null,null,TRUE,FALSE,"[""eth"",""web3"",""net""]","[""100k requests, no more than 1 month, no more than 1 key""]","[""DDoS Protection"",""IP Protection""]","[""Usage analytics""]","[""Global""]",null,null,null,"[{""label"":""Buy"", ""url"":""https://nownodes.io/nodes/flare"", ""type"":""external""}, {""label"":""Docs"", ""url"":""https://nownodes.gitbook.io/documentation-1/nodeapis"", ""type"":""external""}]"
```
ArseniiPetrovich: NowNodes I think we need to offload to RPCs. Other than that looks great!

ArseniiPetrovich (changes requested):

csv_to_json.py
```diff
@@ -0,0 +1,111 @@
...
+        if not os.path.isdir(network_dir_full_path):
+            continue
+
+        result = {}
+
+        # Assumptions:
+        # - Network-specific data is located at: os.path.join(network_dir_full_path, "<category>.csv")
+        # - Provider-specific data is located at: "providers/<category>.csv"
+        # - Category names match both the filename (without extension) and the dictionary key in the result
+        categories = ['rpc', 'indexing', 'oracle', 'bridge', 'explorer', 'faucet', 'analytic', 'wallet', 'devTool']
```
ArseniiPetrovich: What if we have additional categories and then they exists only for specific networks, would that work?

dumikau: We'll need to add the category to the list in this case. The json contains only categories that exist for the network, so adding a new one won't break the script.

networks/flare/rpc.csv
dumikau: Done

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,26 @@
...
+      
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
+      
+      - name: Install dependencies
+        run: pip install --upgrade pip && pip install -r requirements.txt
+
+      - name: Run CSV to JSON
```
ArseniiPetrovich: I think this might be of a security concern. Basically, you run a python script against an unknown input with source code of the execution script known to anyone. What if the in the script we inject some malicious code?

ArseniiPetrovich: I'm not a cybersecurity expert you know, but here is quick analysis on ChatGPT on what to do better here: https://chatgpt.com/share/6899fa71-02c8-8003-9aec-b6e98ec2fadd

I'm worried specifically because this PR will be run against ANY UNVERIFIED code and we have to be secure. I would even add a step maybe with this being triggered only when maintainer adds a label or something along these lines to ensure someone at least checks the PR with their own eyes before running script to ensure it is not something funky out of the box.

ArseniiPetrovich (changes requested):

.github/workflows/validate.yaml
dumikau: Fixed

ArseniiPetrovich (approved):

YanLozovskyi (approved):

PR #232, opened 2025-11-28
feat(validate): slug must be kebab case

### LINKED ISSUE #161: [BUG] Validator bypasses slugs that contain spaces
**Describe the bug**
According to the Kebab case specification - spaces cannot be used in the slug. Our Style Guide mentions that we use Kebab-case. But validator allows slug to have spaces as can be seen in #158 

**To Reproduce**
1. Edit any slug to add space there
2. Commit and run CI
3. CI is mistakingly green

**Expected behavior**
CI not to pass when slug contains space symbol.

dumikau:
Closes #161

### DIFF
--- a/tools/validate.py
+++ b/tools/validate.py
@@ -34,9 +34,10 @@ def rule_slug_unique(data):
 
 def rule_slug_kebab_case(data):
     errors = []
+    KEBAB_CASE_RE = re.compile(r'^[a-z0-9]+(?:-[a-z0-9]+)*$')
     for idx, item in enumerate(data):
         slug = item.get("slug")
-        if not re.match("^[a-z0-9-]+$", slug):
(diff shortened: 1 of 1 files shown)


tools/validate.py
```diff
@@ -32,6 +32,14 @@ def rule_slug_unique(data):
             seen.add(slug)
     return errors
 
+def rule_slug_kebab_case(data):
+    errors = []
+    for idx, item in enumerate(data):
+        slug = item.get("slug")
+        if not re.match("^[a-z0-9-]+$", slug):
```
ArseniiPetrovich: Technically, "-----------" is a match for RegEx, but is it a kebab case or a valid slug?

dumikau: Not anymore

PR #592, opened 2026-02-11
Add verified metrics enrichment stage to JSON generation pipeline

eugene17kotov:
## Summary
Introduces verified SLA metrics enrichment into the JSON generation pipeline.
Verified metrics are now computed during CI and embedded directly into json/<network>.json artifacts.

ClickUp [Task](https://app.clickup.com/t/86af2uptg)

### Changes
- Added `tools/enrich_verified_metrics.py`
- Enrichment updates only apis category
- JSON files are written and committed only if data changed
- Make `validate.py` import-safe by moving csv_to_json import to lazy load
- Use validate.py as single source of truth for schema validation

### Enrichment flow (per network)

- Fetch verified providers (slug, serviceId)
- Batch query SLA subgraph
- Compute:
      - verifiedUptime
      - verifiedLatency
      - verifiedBlocksBehindAvg
- Merge into data["apis"]
- Preserve existing values on SLA failure
- Set verified fields to null if provider is no longer verified

###Required secrets
- VERIFIED_API_TOKEN
- SLA_MONITORING_SUBGRAPH_URL
- GRAPH_API_KEY

### DIFF
--- a/tools/enrich_verified_metrics.py
+++ b/tools/enrich_verified_metrics.py
@@ -7,10 +7,6 @@ import requests
 from jsonschema import Draft202012Validator
 from validate import check_schema_validation  # source of truth
 
-VERIFIED_API_TOKEN_ENV = "VERIFIED_API_TOKEN"
-SLA_MONITORING_SUBGRAPH_URL_ENV = "SLA_MONITORING_SUBGRAPH_URL"
-GRAPH_API_KEY_ENV = "GRAPH_API_KEY"
-
 VERIFIED_CATEGORIES = ["apis"]
 BPS_DENOMINATOR = 10000
 
@@ -278,9 +274,9 @@ def enrich_network_data(
 
 
 def process_all_networks() -> None:
-    verified_api_token = _get_env(VERIFIED_API_TOKEN_ENV)
-    subgraph_url = _get_env(SLA_MONITORING_SUBGRAPH_URL_ENV)
-    graph_api_key = _get_env(GRAPH_API_KEY_ENV)
+    verified_api_token = _get_env(VERIFIED_API_TOKEN)
+    subgraph_url = _get_env(SLA_MONITORING_SUBGRAPH_URL)
+    graph_api_key = _get_env(GRAPH_API_KEY)
 
     if not os.path.isdir("json"):
         print("No 'json' directory found, nothing to enrich")


tools/enrich_verified_metrics.py
```diff
@@ -0,0 +1,382 @@
...
+
+    if not isinstance(metrics_list, list):
+        print(f"[{network}] WARNING: SLA subgraph response has unexpected shape")
+        return None
+
+    metrics_by_id: Dict[str, Any] = {}
+    for m in metrics_list:
+        if not isinstance(m, dict):
+            continue
+        total_proofs = m.get("totalProofs") or 0
```
nkorn1lov: Possible overflow, underlying type in the subgraph schema is `BigInt` for `totalProofs` and `consensusExecutions`, same for other fields that are being converted to `int` in `normalize_metrics`.

tools/enrich_verified_metrics.py
```diff
@@ -0,0 +1,382 @@
...
+        print(f"[{network}] WARNING: verified-providers response has unexpected shape")
+        return None
+
+    normalized: List[Dict[str, Any]] = []
+    for entry in providers:
+        if not isinstance(entry, dict):
+            continue
+        slug = entry.get("slug")
+        service_id = entry.get("serviceId")
+        if isinstance(slug, str) and isinstance(service_id, str) and slug and service_id:
```
nkorn1lov: We use a JSON schema validator in other scripts to validate external data, please refer to https://github.com/Chain-Love/chain-love/blob/json-tools/tools/validate.py.

nkorn1lov (dismissed):

tools/enrich_verified_metrics.py
```diff
@@ -0,0 +1,382 @@
...
+
+    data = payload.get("data") or {}
+    metrics_list = data.get("serviceHealthMetrics") or []
+
+    if not isinstance(metrics_list, list):
+        print(f"[{network}] WARNING: SLA subgraph response has unexpected shape")
+        return None
+
+    metrics_by_id: Dict[str, Any] = {}
+    for m in metrics_list:
```
nkorn1lov: This logic should probably be rewritten using `filter` instead of a loop, similar to the original implementation of verified metrics.

tools/enrich_verified_metrics.py
```diff
@@ -0,0 +1,382 @@
...
+            print(f"[{network}] WARNING: failed to read JSON file: {e}")
+            continue
+
+        providers = fetch_verified_providers(network, verified_api_token)
+        if providers is None:
+            continue
+
+        sla_metrics_by_id: Optional[Dict[str, Any]] = None
+        if len(providers) > 0:
+            service_ids = [p["serviceId"] for p in providers if isinstance(p.get("serviceId"), str)]
```
nkorn1lov: Same validation issue as above, we should try to use schema validation instead of manually checking types of every field of an API response.

tools/enrich_verified_metrics.py
eugene17kotov: Removed redundant isinstance / None checks
Network JSON is now validated against schema.json before and after enrichment.
verified-providers response is validated via a verified_providers.schema.json.
SLA subgraph response is validated via a sla_response.schema.json.

tools/enrich_verified_metrics.py
eugene17kotov: Python 3 int is arbitrary-precision (no fixed 64-bit limit), so it safely handles subgraph BigInt values without overflow

tools/enrich_verified_metrics.py
eugene17kotov: fixed

tools/enrich_verified_metrics.py
eugene17kotov: Removed redundant isinstance / None checks

tools/enrich_verified_metrics.py
```diff
@@ -0,0 +1,350 @@
...
+
+        print(f"[{network}] Processing {path}")
+
+        try:
+            original_data = load_json_file(path)
+        except Exception as e:
+            print(f"[{network}] WARNING: failed to read JSON file: {e}")
+            continue
+
+        if not check_schema_validation(schema_validator=SCHEMA_VALIDATOR, data=original_data):
```
dumikau, line 300: Do we really need this validation here? CSV2JSON workflow runs after a PR that passed the validation is merged into main, we can expect it to be schema compliant at this point.

tools/enrich_verified_metrics.py
```diff
@@ -0,0 +1,350 @@
...
+            )
+
+        enriched_data = enrich_network_data(
+            network=network,
+            original_data=original_data,
+            verified_providers=providers,
+            sla_metrics_by_id=sla_metrics_by_id,
+        )
+
+        if not check_schema_validation(schema_validator=SCHEMA_VALIDATOR, data=enriched_data):
```
dumikau, line 325: This validation is fine though. We've changed the data, we need to test if it's still compliant.

dumikau: You could as well import a similar function from `validate.py` just so we have only one source of truth for validation process.

tools/enrich_verified_metrics.py
eugene17kotov: We also use this in the second `verified-refresh` workflow, which runs on cron. But I agree that the extra check here might be too much guardness.

tools/enrich_verified_metrics.py
eugene17kotov: I completely agree with this. We just import it from `validate.py`

Image

PR #676, opened 2026-02-26
feat(meta): add group categorization to columns in JSON schema

eugene17kotov:
- Updated `columns.json` to include a new "group" field for various columns, categorizing them into relevant groups such as "Identity", "Highlights", "Links", "Capabilities", and "Pricing & SLAs".
- Enhanced `schema.json` to support new enum values for `planType`, `apiType`, `technology`, and `historicalData`, improving validation and structure of the data model.

### DIFF
--- a/meta/columns.json
+++ b/meta/columns.json
@@ -8,7 +8,7 @@
     "sorting": null,
     "pinning": null,
     "cellType": null,
-    "group": "Identity"
+    "group": "identity"
   },
   "starred": {
     "key": "starred",
@@ -19,7 +19,7 @@
     "sorting": null,
     "pinning": null,
     "cellType": null,
-    "group": "Highlights"
+    "group": "highlights"
   },
   "provider": {
     "key": "provider",
@@ -30,7 +30,7 @@
     "sorting": "string",
     "pinning": "left",
     "cellType": "provider",
-    "group": "Identity"
+    "group": "identity"
   },
   "actionButtons": {
     "key": "actionButtons",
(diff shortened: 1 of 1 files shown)


tools/schema.json
```diff
@@ -35,8 +35,18 @@
...
+        "planType": {
+          "oneOf": [
+            { "type": "string", "enum": ["Free", "Pay-as-you-go", "Enterprise", "Dedicated", "Tiered", "One-time payment"] },
+            { "type": "string" }
+          ]
+        },
+        "apiType": {
+          "oneOf": [
+            { "type": "string", "enum": ["RPC", "Indexed Data"] },
+            { "type": "string" }
```
dumikau: `oneOf` means that the value must match exactly one of the listed schemas. If the value is `"RPC"`, it matches both schemas listed, not just one of them, and it will cause error like `"RPC" is valid under each of { "type": "string" }, { "type": "string", "enum": ["RPC", "Indexed Data"] }`. Please leave only the schema with enum here and in other places with the same problem.

eugene17kotov: You’re right about the oneOf conflict; since both schemas match a string, it causes a validation error.

However, limiting this to a strict enum won't work for our logic, as the field must support custom user input (arbitrary strings) alongside the predefined options.

To keep it valid yet flexible, I suggest using type: "string" with examples. This allows the UI to show the select options while permitting any custom string:
```
JSON
"planType": {
  "type": "string",
  "examples": [
    "Free", "Pay-as-you-go", "Enterprise", "Dedicated", "Tiered", "One-time payment"
  ]
}
```
What do you think?

ArseniiPetrovich (approved):

PR #686, opened 2026-02-27
images

harunsulaiman:
## Type of change
- [ ] Add data rows
- [ ] Update data rows
- [ ] Remove data rows
- [ ] Schema change 

## Scope
- Networks affected: 
- Categories affected: 

## Links
- Related issue(s)/ DB Improvement Proposal (if schema-related): 

## Validation checklist

- [ ] **I followed the Style Guide and Column Definitions. I'm aware of what is `!provider` syntax, and that entities in `/networks` sub-folders inherits records from `/providers` folder**
  - Style Guide: https://github.com/Chain-Love/chain-love/wiki/Style-Guide
  - Column Definitions: https://github.com/Chain-Love/chain-love/wiki
- [ ] **I personally opened and verified every new link I'm adding. I can confirm, that all the links I'm adding are valid.**
- [ ] **If I added new entries - I personally confirmed that the provider I'm adding (modifying) currently supports the adjusted network(s). I've also verified that value in every cell I'm changing is correct according to my best understanding**

## Optional
- Rewards address (for data patching rewards):
- Twitter (X) post link (for +10% of rewards to this PR):


geraBarboni (changes requested):
Hey @harunsulaiman! Could you change the images to PNG with a transparent background? Currently, both have a white background 😅.
And for the 3xpl one, there are two options:

Use one that doesn't have the black border.
Add a margin between the edge of the image (10-20px) and the black border.

harunsulaiman:
@geraBarboni images are in png.    and the options you are providing means i have to edit it using image editing softwares.    thats like,  big work.   

the images i used are the official logos in the required size.     i think personally, backgrounds don't matter since it is the official logo used the same way in their official website.


@ArseniiPetrovich

ArseniiPetrovich:
The background removal are quite easy, really. It took me like 30 seconds to google and use a completely free solution that removes background automatically, so that's that (https://www.remove.bg/pl/upload)
ambire-removebg-preview
As for the 3xpl one - it is more complicated, I guess? @geraBarboni, can we have some simple solution that can be applied by our contributors to adapt pictures? I mean, background removers are well known, easy to use and free, so should be all good, but if you need to remove like a black color from a logo... I'm not quite sure it is on the same level of simplicity? I tried to go with colors inversion, and then background remover, but it keeps removing the center of the picture for me for the 3xpl.
Lets have a discussion tomorrow to identify a more-less simple way to contribute, and a bit clearer guideline, as I myself wouldn't get as well that the black picture wouldn't fit, as it is not a background, but a part of their logo? I guess? Yeah, lets discuss anyways.

harunsulaiman:
> @harunsulaiman so, the consensus here is that it is recommended to remove background where possible. @geraBarboni will provide clearer instructions later this week, but for 3xpl, for example, we agreed to keep everything as is, as black color seems to be a part of their logo, and if we remove that - it will be simply hardly visible. For ambire - and for similar logos where white background is not "mandatory" - we recommend removing it with background removal tool.

ok noted

PR #1121, opened 2026-03-25
chore(ci): fetch ready to use agents from app.chain.love

dumikau:


USS-Supervisor (dismissed):
Requesting changes due to functional regressions in :\n\n1) Schema path resolution changed from file-relative lookup to , which is brittle and can fail depending on working directory.\n2) The new loop now raises hard  on missing  and aborts the whole run, replacing the previous per-network skip behavior.\n\nPlease restore robust path handling and non-fatal per-network error handling before merge.

Igor2698 (dismissed):

Igor2698 (approved):

eugene17kotov (approved):

PR #1131, opened 2026-03-26
feat: add create_ndjson.py script

dumikau:
In release, in addition to existing files, this will create tar archives following this pattern: `<network_name>-ndjson.tar`

Each such archive will contain multiple `<category_name>.ndjson` files.

### DIFF
--- a/tools/create_ndjson.py
+++ b/tools/create_ndjson.py
@@ -1,0 +1,79 @@
+from __future__ import annotations
+
+import os
+import json
+import tarfile
+import io
+from typing import Dict, List, Union, cast
+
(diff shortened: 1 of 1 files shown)


USS-Supervisor (commented):
Re-review complete.

I did not find a data-integrity blocker in the added script itself, but I can't approve this cycle because the current review-cycle `link-check` result is `missing`.

Also, the PR currently has a failing `validate` check (`requirements.txt` missing in that workflow run), so please re-check CI configuration/outcome before merge.

ArseniiPetrovich (approved):

Issue #2481, opened 2026-08-08
[DBIP] explorers is the only category that names the common price column `pricing`


EazyHood:
### Proposal type

Modify column

### Affected scope (files/folders/chains)

references/offers/explorers.csv; listings/all-networks/explorers.csv; listings/specific-networks/*/explorers.csv; wiki `Explorers` page

### Motivation / problem statement

`price` is a **common column**: it has its own entry in `Common-Columns` and eleven of the fourteen
offer categories use that exact name. `explorers` is the only category that stores the same thing
under a different name, `pricing`.

The two definitions describe the same field:

| | `price` (Common-Columns) | `pricing` (Explorers) |
|---|---|---|
| definition | Pricing model | Access tier for advanced features/APIs |
| format | text | text |
| examples | `$0`, `$X/month`, `Custom` | `Free`, `$100/mo` |

And the values in the file are not distinguishable from any other category's. All 60 rows of
`references/offers/explorers.csv` carry a value, and they read exactly like the others:

```
explorers.pricing   Free | $495/mo | $995/mo | $1995/mo | Developer ($54/mo - 10k calls/day)
services.price      $0   | $19/mo  | $199/mo | $250+    | $0.02/Gas fee
sdks.price          $0   | $20/mo  | $100/mo | $149/mo  | $2.8K/mo
```

Who benefits: anyone reading or querying across categories. "Show me the free tools" or "sort by
price" has to special-case one file out of fourteen today, and a contributor who has filled in
`price` a dozen times will reach for it in `explorers.csv` and find it is not there.

### Detailed proposal

- Category/table: explorers
- Column name: `pricing` -> `price`
- Change type: modify (rename)
- New/updated definition: the existing `Common-Columns` entry for `price` - "Pricing model" -
  applies unchanged. The Explorers page can drop its own `pricing` entry, since the column would
  then be a common one.
- Value type and allowed values: text. **Unchanged** - this is a header rename only. No cell
  needs to be rewritten, in `references/offers/explorers.csv` or in any listing.
- Examples: `$0`, `Free`, `$495/mo`
- Notes:
  - **`apis` is deliberately not part of this.** It has `accessPrice` and `queryPrice` instead of
    `price`, and those are two genuinely different quantities - "monthly access price" and "price
    per 1M requests (or CUs)" - each with its own wiki entry. That is a considered split, not a
    naming drift, and I am not proposing to touch it.
  - `oracles` and `ramps` have no price column at all. That is a different question and I have
[...]

PR #2506, opened 2026-08-08
feat(explorers): add Ethereum and Arbitrum explorer options

cadfix-lab-cbccfc1d:
## Summary

Add six current explorer products with verified Ethereum and Arbitrum availability. This fills gaps for blob data, staking analytics, transaction debugging, human-readable multichain activity, and hosted explorer infrastructure.

## Type of change

- [x] Add data rows
- [ ] Update data rows
- [ ] Remove data rows
- [ ] Schema change
- [ ] Documentation/metadata only

## Scope

- Networks affected: Ethereum and Arbitrum
- Categories affected: explorers
- Additional notes: adds five provider references, six explorer offers, six Ethereum mainnet listings, and three Arbitrum One listings

## Links

- Related issue / DBIP: not applicable; no schema change

## Validation checklist

- [x] I followed the Style Guide and Column Definitions and used offer inheritance for network listings.
- [x] I opened and verified every new link in this change.
- [x] I confirmed the providers currently support each listed network and checked every populated field against official sources.
- [x] This is not a blind AI-generated submission; the data was individually researched, deduplicated, and validated.

## Verification

- `python validate_csv.py` - all checks passed
- `python csv_to_json.py` - generated successfully with the repository's six pre-existing missing-chain warnings
- `python validate.py` - all generated network JSON files passed schema validation

## Official sources

- Altscan / EthScan: https://altscan.io/ and https://ethscan.io/
- Blobscan: https://blobscan.com/ and https://docs.blobscan.com/
- Ethernal: https://tryethernal.com/ and https://doc.tryethernal.com/
- Moca Explorer: https://moca.app/explorer and https://docs.moca.app/
- Blocksec Phalcon Explorer: https://blocksec.com/phalcon/explorer and https://docs.blocksec.com/phalcon/phalcon-explorer
- Rated Explorer: https://explorer.rated.network/ and https://docs.rated.network/documentation/explorer/introduction

## Grant estimate

- 20 added rows
- 136 populated cells
- estimated reward: $8.16 at $0.06 per accepted populated cell

## Optional

- Rewards address: `0x2B098B360c86a22474856872f0B8F875D9D8B05E`
- Twitter (X) post link: none

PR #2512, opened 2026-08-09
Rename explorers pricing to common price column (DBIP #2481)

elevasyncsolutions-jpg:
Explorers is the only one of 14 offer categories storing the common Pricing-model field as `pricing` instead of `price`. Header-only rename: `references/offers/explorers.csv` + 77 network listing files, 0 cell changes, 0 network-specific pages affected.

Paired tools PR (json-tools): schema explorers $def + columns meta.


elevasyncsolutions-jpg:
These two PRs are schema-coupled and must land as a pair:

1. **#2513 (json-tools)** — schema `$defs.explorers` requires `price` instead of `pricing`
2. **#2512 (main data)** — renames the header so the data matches the new schema

#2512's `Generate JSON and validate (isolated)` is red **only because the current canonical schema still requires `pricing`** (validated locally: `'pricing' is a required property` at `#/explorers/59`). Merge #2513 first (or both together) and #2512 goes green. Note for other open PRs touching explorers CSVs (#2506, #2514): they'll need a `pricing`->`price` header rebase after this lands.

EazyHood:
Thanks for picking this up so quickly - the rename itself is exactly what I meant, and splitting
the tools change into #2513 was the right call.

One thing to look at before this lands. The description says "No cell values changed", and for the
explorers files that holds, but the diff also touches **six data rows in three files that have no
`pricing` column at all**:

| file | slug | `offer` before | after |
|---|---|---|---|
| `listings/all-networks/storages.csv` | `thread-db` | `Thread DB` | *(empty)* |
| `listings/specific-networks/avalanche/sdks.csv` | `avacloud-sdk` | `SDK` | *(empty)* |
| | `avalanche-tooling-sdk-go` | `SDK` | *(empty)* |
| | `avalanche-wallet-sdk` | `SDK` | *(empty)* |
| | `avalanchejs` | `SDK` | *(empty)* |
| `listings/specific-networks/lens/faucets.csv` | `lenscan-testnet-faucet` | `Lenscan` | *(empty)* |

Column 3 is `offer` in all three. That is why the diff is +83/-83 across 80 files rather than the
+80/-80 a pure header rename would produce.

The rule behind it is exact, which should make it easy to find: in each of those three files, the
rows carrying a **literal** `offer` value - one that is not a `!offer:<slug>` reference - are
**precisely** the rows being emptied. Six of six, with no other row touched:

```
storages.csv (all-networks)  20 rows, 1 literal offer  -> that 1 blanked
avalanche/sdks.csv           61 rows, 4 literal offers -> those 4 blanked
lens/faucets.csv              3 rows, 1 literal offer  -> that 1 blanked
```

So it looks like whatever round-trips the CSVs drops `offer` when it is not a reference, rather
than anything to do with `pricing`.

`AGENTS.md` does say listings should use `!offer:<slug>` in that column, so those six values may
well be non-conformant - but converting them is a different change from renaming a column, and
blanking them loses the name rather than replacing it. My suggestion would be to restore the six
cells here and, if they should be references, do that conversion in its own PR where a reviewer can
see it.

Nothing else in the diff surprised me: of the 80 files, the **77** outside the table above are
each a single-line header rename and nothing more. (Correcting my own arithmetic - I first wrote 79
there, which does not square with naming three files as exceptions.)

USS-Supervisor:
Hi @elevasyncsolutions-jpg - this PR looks stale, and review should wait because required CI is not passing (`Generate JSON and validate (isolated)=fail/FAILURE`). Could you please fix CI and update the PR when it is ready for review?

USS-Supervisor:
Hi @elevasyncsolutions-jpg - this PR is still stale, and review should wait until required CI is green. The current blocker is Generate JSON and validate (isolated)=COMPLETED/FAILURE, required validation missing/not successful. Could you please fix CI and update the PR when it is ready for review?

USS-Supervisor:
@elevasyncsolutions-jpg closing this for now because the current blocker is CI not passing, the next step was with the PR creator, and the PR stayed unattended for more than 7 full days after the supervisor ping on 2026-08-17T14:38:09Z. This can be reopened or recreated when ready.

PR #2637, opened 2026-08-11
data(providers): fill missing contact and social fields (batch 3)

EazyHood:
Fills empty contact and social fields in `references/providers/providers.csv` from each provider's
own site and documentation. No existing values were changed; only empty cells were populated.

## Rewards address
0x06f44f4839fd5df4f4670036d028b29dec939363


USS-Supervisor (changes requested):
Verdict: REQUEST_CHANGES
Risk: MEDIUM
Summary: The provider CSV shape and row set stay stable, but one newly populated social field points Ava Labs at an unrelated Telegram account. The current-cycle link-check run completed, but its bot report still listed 2 URL errors in this touched file, so approval would not be available in this cycle anyway.
Findings:
- MEDIUM, `references/providers/providers.csv` row 62: `avalabs` now sets `telegram` to `avacoin_official`, but `https://t.me/avacoin_official` identifies itself as a different account rather than Ava Labs/Avalanche. This attaches the Ava Labs provider row to an unrelated Telegram channel. Please remove this value or replace it only with a source-backed official Ava Labs/Avalanche Telegram handle.
Confidence: HIGH

EazyHood:
Fixed. The branch is now rebased onto current main, so the four provider rows that were reported as
removed — cidgravity, fil-one, filecoin-onchain-cloud and palmera — are present and byte-identical
to main. The diff is now symmetric and contains only the filled contact cells.

The flagged social handles have been cleared rather than guessed. I also re-checked every value this
PR adds rather than only the ones you flagged: all X handles and Discord invites resolve, and for
Telegram I compared each channel's own title against the provider instead of only checking that the
link loads. That surfaced one more bad handle you had not flagged — `blueprint` pointed at a channel
titled "QDev Studio" — which is now cleared too.

USS-Supervisor:
@eugene17kotov This PR has passing required CI and is waiting on reviewer action. Could you please take a look when you have a chance?

EazyHood:
The changes requested here have been addressed in the latest push. Could you take another look when you get a chance?

EazyHood:
Superseded by #3250, which rebuilds all five of these against current main as a single change.

`providers.csv` has taken 70+ commits since 11 August, so these were re-conflicting faster than they could be rebased - and merging any one of them re-conflicted the other four, since all five edit the same file. #3250 applies every cell from all five that is still empty on main (463 cells across 317 rows), with zero disagreements between them.

Closing to keep the queue clean. The branch stays on my fork if anything here needs recovering.

PR #2727, opened 2026-08-12
fix(providers): normalize github to organizations

### LINKED ISSUE #2460: [DBIP] The `github` column holds two different kinds of value
## Problem

`references/providers/providers.csv` has 413 rows with a non-empty `github` value.
They are not the same kind of thing:

  320 (77.5%)  organization      1inch, 2miners, 0xRPC
   93 (22.5%)  organization/repo 0xcert/framework, Vid201/aa-bundler

`account-smart-wallet` is `base/account-sdk`, which is not a repository owned by
that provider at all.

The `x` column does not have this problem: all 433 non-empty values are handles.

## Why it matters

A consumer of the CSV cannot resolve the column to a URL without guessing. Prefixing
`https://github.com/` yields an organization page for 320 rows and a repository page
for 93, and for `account-smart-wallet` it yields someone else's repository.

It also makes the column impossible to fill correctly: a contributor looking at
existing rows sees both conventions and cannot tell which one is intended.

## Proposal

Define `github` as organization-level only, and move the 93 repository-valued rows
to the organization that owns them.

If repository-level detail is wanted, that is a separate column (`repo`), which this
DBIP does not propose adding.

## Scope

93 rows to correct. I can submit them as a PR once the convention is confirmed.

bcabreraike-cmyk:
## Summary

Normalizes the providers github field to organization-only values by removing repository suffixes after slash. No provider metadata other than github is changed.

## Type of change
- [x] Update data rows

## Scope
- Networks affected: global provider metadata.
- Category affected: providers.
- Related DBIP: Closes #2460

## Rationale
The current field mixes organization values with repository paths. A consumer cannot resolve the field consistently without guessing whether it should open an organization or repository page. This patch follows the DBIP organization-level convention.

## Validation
- 713 provider rows were compared before and after.
- 97 slash-containing values were normalized to the organization segment.
- 0 slash-containing github values remain.
- All non-github fields and all existing organization-only github values were preserved.

## Rewards address
0xf04865D955D5f698951c3eBFdb6654EF5Ab17ac (Ethereum Mainnet ERC20)


USS-Supervisor (dismissed):
Verdict: APPROVE
Risk: LOW
Summary: Required validation is passing and the current review-cycle link-check completed successfully. I did not find a blocking Chain.Love data-integrity issue in the visible diff.
Findings: None blocking.
Confidence: MEDIUM

eugene17kotov (changes requested):
Thanks for working on provider consistency. I don't think this bulk normalization is safe to merge in its current form.

The `github` field is not restricted to an organization/user name. According to the repository rules, it stores the value after `github.com/`, so both `owner` and `owner/repo` are valid depending on the provider.

For many product-level providers, the existing repository path is actually more accurate than the proposed organization-only value. For example, `NomicFoundation/hardhat`, `ethers-io/ethers.js`, `paradigmxyz/reth`, and `dethcrypto/TypeChain` point directly to the canonical projects represented by those provider rows; replacing them with only the repository owner loses useful provider-specific information.

There may be individual rows where the current repository is incorrect or unrelated to the provider — those should absolutely be fixed, but they need to be verified case by case rather than by stripping the repository component globally.

Please revert the bulk `owner/repo` → `owner` changes and keep only individually verified corrections where the organization/user page is genuinely a better provider-level GitHub entry point. Please also avoid unrelated quote-only URL rewrites, as they add a large amount of diff noise without changing the data.

eugene17kotov (approved):

USS-Supervisor (approved):
Verdict: APPROVE
Risk: LOW
Summary: The current head has an empty diff relative to main, so the earlier unsafe provider GitHub normalization is no longer present. Required validation is passing, and the current review-cycle link-check run 32071532426 completed successfully with the bot report showing 0 errors.
Findings: None blocking.
Confidence: HIGH

USS-Compensator (commented):
## Suggested compensation for PR #2727

Contributor: @bcabreraike-cmyk

This is a suggested compensation only, based on the current interpretation of the reward program. Final compensation may be altered by a human reviewer.

Basis checked:
- Discussion #41: 0.06 USDC per improved non-null cell, 0.10 USDC per added image, the 200-cell limit, creation-date bounty-network eligibility, the 10% qualifying X-post bonus, and program-defined 30% quality reductions.
- Discussion #839 (2026-09-07 digest): current category modifiers; provider metadata remains at x1.00.
- Deterministic command: `python3 compensator.py --repo Chain-Love/chain-love --pr 2727 --no-header --key-config compensator.key-config.json --json`

Calculation:
- Raw deterministic author-owned non-null cells: 0.
- Eligible raw contribution count used: 0 cells.
- Eligibility result: The final PR diff is empty.
- Added-image reward: 0. No payable added image was found.
- Quality deductions: 0. The recorded change request was from USS-Supervisor, which is excluded from the Discussion #41 quality-deduction count.
- X-post bonus: 0. No qualifying post link was established.
- Cap: 200 billable cells; not reached by the eligible raw count.

Final suggested compensation: **0.00 USDC**.

Human reviewer note: this calculation is advisory and may be adjusted after human review of eligibility, attribution, or program interpretation.

PR #3309, opened 2026-08-30
Fill verified provider links for Filecoin-listed services

gumingxiang:
## Summary

Fill 18 previously blank provider metadata cells across 12 existing providers that are already referenced by Filecoin listings. The patch adds only current documentation, community, LinkedIn, Telegram, and support-contact values surfaced directly by each provider's official website.

No provider names, descriptions, offers, listings, network-support claims, or pricing fields are changed.

## Type of change

- [x] Update data rows
- [ ] Add data rows
- [ ] Remove data rows
- [ ] Schema change
- [ ] Documentation/metadata only

## Scope

- Networks affected: Filecoin (provider records are global; every edited provider already resolves from at least one Filecoin listing)
- Categories affected: providers

## Official sources

- Atomic: https://atomicwallet.io/
- Beosin: https://beosin.com/
- BlockEden: https://blockeden.xyz/
- Blockscout: https://www.blockscout.com/
- GLIF: https://www.glif.io/
- Pinata: https://pinata.cloud/
- Pyth Network: https://www.pyth.network/
- Squid: https://www.squidrouter.com/
- thirdweb: https://thirdweb.com/
- Trust Wallet: https://trustwallet.com/
- wagmi: https://wagmi.sh/
- Zondax: https://zondax.ch/

Each source page was fetched successfully and contained the exact link or contact value added here. Direct HTTP links were also fetched successfully before this PR was opened.

## Validation

- Parsed all 720 provider rows as 14-column CSV.
- Confirmed all 18 target values landed in the intended columns.
- Confirmed provider slugs remain unique and A-Z sorted.
- Confirmed every edited provider is already referenced by a Filecoin listing.
- Searched the complete diffs of all 471 open pull requests for these provider rows; no competing edits were found.
- `git diff --check` passes.

## Rewards

Rewards address: `0x5a6223dD34ae05F3DD6f6AEee7a5a683376Cc5d5`

PR #3250, opened 2026-08-27
data(providers): consolidate 463 verified contact and social fields from five conflicting PRs

EazyHood:
## Summary

Rebuilt against current main from five of my own PRs that have been sitting in conflict: #2635,
#2636, #2637, #2638 and #2993.

`providers.csv` has taken 70 commits since 11 August, so rebasing them one at a time only puts
them back in conflict a couple of days later - and merging any one of them re-conflicts the other
four, since all five edit the same file. Consolidating into one branch against today's main is the
only way out of that loop.

This applies every cell from those five that is still empty on main today, and nothing else:

| column | cells |
|---|---|
| `supportEmail` | 164 |
| `discord` | 108 |
| `x` | 65 |
| `telegram` | 60 |
| `docs` | 39 |
| `github` | 10 |
| `website` | 9 |
| `linkedin` | 8 |

463 cells across 317 provider rows. By source: #2636 contributes 166, #2993 101, #2637 91, #2635
73, #2638 32.

**No existing value is overwritten.** The change only fills cells that are blank on main, and the
five PRs proposed no conflicting values between them - zero disagreements across all 463.

Verified row by row after the edit: 718 rows before and after, every changed row keeps its slug,
and no changed cell had a prior value. `validate_csv.py`, `csv_to_json.py` and `validate.py` all
pass; the six `csv_to_json` warnings about camino, lightning and zilliqa are pre-existing and
appear identically without this change.

If this lands, #2635, #2636, #2637, #2638 and #2993 can be closed as superseded. I will close them
myself once it is merged.

## Type of change
- [x] Update data rows

## Scope
- Networks affected: global (references/providers/providers.csv)
- Categories affected: none - provider metadata only

- Additional notes: edits were made line by line rather than by reserializing the CSV, so quoting
elsewhere in the file is untouched, and line endings are unchanged (the file stays LF throughout).
The diff is 317 insertions and 317 deletions - one line per touched provider row.

## Links
- Supersedes #2635, #2636, #2637, #2638, #2993

## Validation checklist

- [x] **I followed the Style Guide and Column Definitions. I'm aware of what is `!provider` syntax, and that entities in `/networks` sub-folders inherits records from `/providers` folder**
- [x] **I personally opened and verified every new link I'm adding. I can confirm, that all the links I'm adding are valid.**
- [x] **If I added new entries - I personally confirmed that the provider I'm adding (modifying) currently supports the adjusted network(s). I've also verified that value in every cell I'm changing is correct according to my best understanding**
[...]

PR #3218, opened 2026-08-25
data(ethereum): add Foundation staking and core libraries

edras-create:
## Summary

Adds ten source-backed Ethereum Foundation staking, networking, forecasting, privacy, simulation, cryptography, utility, hashing, typing, and specification projects: Staking Launchpad, Node Crawler, Forkcast, py-ecc, Ethshadow, eth-utils, Kohaku, eth-hash, eth-typing, and ethereum-types.

## Type of change
- [x] Add data rows
- [ ] Update data rows
- [ ] Remove data rows
- [ ] Schema change
- [ ] Documentation/metadata only

## Scope
- Network affected: Ethereum
- Categories affected: analytics, platforms, SDKs, providers
- 21 added CSV rows across 7 files
- 169 new non-empty cells; below the 200-cell and 100-line limits

## Sources
- https://github.com/ethereum/staking-launchpad
- https://github.com/ethereum/node-crawler
- https://github.com/ethereum/forkcast
- https://github.com/ethereum/py_ecc
- https://github.com/ethereum/ethshadow
- https://github.com/ethereum/eth-utils
- https://github.com/ethereum/kohaku
- https://github.com/ethereum/eth-hash
- https://github.com/ethereum/eth-typing
- https://github.com/ethereum/ethereum-types

## Duplicate and quality audit
- Checked current main and all open PRs for exact repository URLs and semantic overlap immediately before preparation.
- Excluded discontinued, unmaintained, archived, duplicate, already listed, and already claimed projects.
- Added one canonical Ethereum Foundation provider row and used conservative first-party descriptions.
- The display name py-ecc avoids an unclosed Markdown token while matching the published package name.

## Validation
- Official validate_csv.py: PASS
- Full CSV-to-JSON generation: PASS
- Full generated JSON schema validation: PASS
- Slug uniqueness, canonical !offer: references, CSV widths, ordering, and git diff --check: PASS
- Diff: 21 additions, 0 deletions

## Validation checklist
- [x] I followed the Style Guide and Column Definitions.
- [x] I opened and verified every included source link.
- [x] I confirmed every tool supports Ethereum and reviewed every changed cell.
- [x] This is not a blind AI-generated submission.

## Bounty accounting
- 169 potentially eligible non-empty cells
- Base nominal amount: **10.14 USDC** at 0.06 USDC/cell, before category multipliers and subject to maintainer approval

## Optional
[...]

PR #3211, opened 2026-08-25
data(algorand): add Tinyman official Python and JavaScript SDKs

stmr:
## Summary

Adds Tinyman's official Algorand AMM SDKs that were missing from the Algorand SDK table:

- provider identity + 256x256 transparent PNG from the official Tinyman app icon
- canonical Python SDK offer (`tinyman-py-sdk`)
- canonical JavaScript SDK offer (`tinyman-js-sdk`)
- Algorand listings via `!offer:tinyman-py-sdk` and `!offer:tinyman-js-sdk`

Provider display name is **Tinyman** (organization), not a product string, so the offers render as "Tinyman Python SDK" and "Tinyman JavaScript SDK" per Common-Columns.

Official docs list both SDKs as started and maintained by the Tinyman core team.

## Sources

- App: https://app.tinyman.org
- Website: https://tinyman.org/
- Docs overview: https://docs.tinyman.org/
- Official SDKs: https://docs.tinyman.org/v2-integration/official-sdks
- Python SDK / MIT: https://github.com/tinymanorg/tinyman-py-sdk
- JavaScript/TypeScript SDK / MIT / `@tinymanorg/tinyman-js-sdk`: https://github.com/tinymanorg/tinyman-js-sdk
- Official icon: https://github.com/tinymanorg/brand/blob/master/App_Icon/App_Icon_256px.png

## Type of change

- [x] Add data rows
- [ ] Update data rows
- [ ] Remove data rows
- [ ] Schema change
- [ ] Documentation/metadata only

## Scope

- Networks affected: Algorand
- Categories affected: sdks, providers
- Rows added: 1 provider, 2 canonical offers, 2 Algorand listings
- Image added: `references/providers/images/tinyman.png` (256x256 RGBA PNG from the official app icon)

## Verification

- `python3 git-hooks/pre-commit.py` passed locally after keeping only these four files
- Live URL checks: `https://docs.tinyman.org/v2-integration/official-sdks` 200, `https://github.com/tinymanorg/tinyman-py-sdk` 200, `https://github.com/tinymanorg/tinyman-js-sdk` 200

## Payout

Ethereum payout address for USDC/USDT: `0x10DFE6771131BEc830A7a44D7Be45Ced0741b2b3`

PR #3323, opened 2026-08-31
data(providers): fill 40 missing link cells for 28 providers (batch 1 of 4)

EazyHood:
## Summary

Fills **40 empty cells** across **28 providers** in `references/providers/providers.csv` (`1rpc` … `forge`). No rows added or removed, and no cell that already had a value was changed.

| column | cells filled |
|---|---:|
| `github` | 8 |
| `docs` | 7 |
| `x` | 13 |
| `discord` | 7 |
| `telegram` | 5 |

One of four batches (#3323, #3324, #3325, #3326) covering different providers; no cell appears in more than one of them. **Rebased on current `main`** — the earlier version conflicted after #3211, #3218, #2814 and #3309 touched this file, which is also why no workflow ran on it (GitHub does not trigger `pull_request` workflows on a conflicting PR).

## Type of change
- [ ] Add data rows
- [x] Update data rows
- [ ] Remove data rows
- [ ] Schema change

## Scope
- Networks affected: global (provider identities, not listings)
- Categories affected: `references/providers`

**Overlap with #3250, stated up front.** #3250 fills 450 provider cells and 15 of the cells in this PR are also in it — **with the same value in every case**; where the two disagreed, the cell was removed from here. Whichever lands second rebases and those cells become a no-op in it, so nothing is claimed twice. They are kept here because #3250 is large and has been cycling on link-check noise since 27 August, and these batches are small enough to clear a review cycle on their own.

**Where the values come from.** Two first-party sources, never a guess from the slug: the provider's own website (outbound links and `twitter:site` meta tag), and the provider's own GitHub organisation profile (`twitter_username`), plus the provider's own `docs.`/`developers.` subdomain accepted only when the URL after redirects stays on the provider's registrable domain. A value was written only when unambiguous; where two plausible candidates existed, the cell was left empty. That is why these four PRs fill ~160 cells and not the ~650 that are empty.

**Accuracy, measured before writing anything**, by running the same extraction against providers whose value is *already* recorded:

```
github  64/67   agree   95.5%
x      114/117  agree   97.4%   (from the provider's site)
x      171/175  agree   97.7%   (from the GitHub org profile, after a name-similarity filter)
```

A third source was measured and **rejected**: the GitHub profile's `blog` field agrees with the recorded `website` only 80.7% of the time, so no `website` cell is touched.

**Link checking.** Every link in every touched row — the cells added here *and* the `website`/`docs` values already present in those rows — was requested with TLS verification on and compared against the accept list this repo's link-check uses (`200,202,204,400,401,403,405,429`): across the four batches, 523 links, 0 failures. Three rows that could have been filled were **left out** because they carry a link that was already in the database and fails today (`swing`: expired certificate and a docs domain that no longer resolves; `polyzoa`: 404; `learnweb3`: 503) — since the check reads whole rows, touching them reports those failures against this PR. No `linkedin` cell is included, for the reason in #3347.

**Changed since the first review on this PR:** `core-dao`'s `x` was `name` — Core DAO's own page carries `<meta name="twitter:site" content="@name">`, an unfilled template value that my extractor took at face value. It is now `Coredao_Org`, which is what the same page links to and which returns 200.

**Style Guide conformance.** `docs` as a full quoted URL; `x`, `github`, `discord`, `telegram` as value-after-domain only; `discord` as the bare invite code, matching 125 of the 136 existing values. Row order untouched, so the file stays A–Z by `slug`.

## Links
- Related issue(s) / DB Improvement Proposal: none — data population, no schema change. See #3347 for why `linkedin` is excluded.

## Validation checklist

  - To be exact about what "verified" means here: every link was checked with an HTTP request as described above, not opened one by one in a browser. Every value in this PR returned an accepted status; everything that did not was removed before submitting.

## Optional
- Rewards address (for data patching rewards): 0x06f44f4839fd5df4f4670036d028b29dec939363
- Twitter (X) post link (for +10% of rewards to this PR): https://x.com/del_mejia95726/status/2095272702319743473


USS-Supervisor (changes requested):
Verdict: REQUEST_CHANGES
Risk: MEDIUM
Summary: The provider CSV remains structurally valid, but one newly filled provider social field is unsupported by the provider's own site.
Findings: MEDIUM - `references/providers/providers.csv`, row `core-dao`, field `x`: the PR sets the X handle to `name`, which appears to be leaked metadata rather than Core DAO's official X account. Core DAO's own website links to `twitter.com/Coredao_Org`. Please replace this with `Coredao_Org` or leave the field blank if you cannot confirm the official handle.
Confidence: HIGH

Required validation passed. Current-cycle link-check run `33375666339` completed with workflow success but its report still showed 1 URL error, so this review cannot approve the PR in this cycle.

EazyHood:
One thing worth flagging before the link-check runs on this, because it is likely to report `x.com` URLs as errors and they will not be broken links.

X throttles GitHub-hosted runners. In the most recent link-check run on #2977 — 2663 links checked — **25 of the 42 errors were on `x.com` or `linkedin.com`**, including:

```
[ERROR] <https://x.com/stripe>
[ERROR] <https://x.com/SuiNetwork>
[ERROR] <https://x.com/ton_blockchain>
```

Those accounts are live; all three return 200 from an ordinary connection. The same handle comes back as `ERROR`, `404` or `500` between runs depending on throttling, which is why #2977, #2978 and #2979 have been cycling on this since 16 August.

I have already removed every `linkedin` value from these four PRs for that reason — 40 cells I had verified — and opened #3347 proposing a configuration fix.

For the `x` values that remain here: each one was taken from the provider's own website or its GitHub organisation profile, and each returned 200 on `x.com` at submission time. I have not removed them because they are correct and because doing so would leave the column permanently unfillable. **If the check flags a specific handle in this PR, say which one and I will drop that cell in the same hour** — I would just rather not delete 75 verified values pre-emptively against an intermittent signal.

Everything else in these rows was re-checked against your accept list (`200,202,204,400,401,403,405,429`): 523 links across the four batches, 0 failures.

EazyHood:
On the `core-dao` finding: you are right, and the cause is worth a line. Core DAO's own page ships `<meta name="twitter:site" content="@name">` — a Next.js template value nobody filled in — and my extractor took the meta tag at face value. The same page links `twitter.com/Coredao_Org`, which is what the cell now holds; it returns 200.

While fixing that I rebased all four batches on current `main` (the earlier heads had conflicted after #3211, #3218, #2814 and #3309, which is also why no workflow ran on them), removed the `swing`, `polyzoa` and `learnweb3` rows whose pre-existing links fail the check, removed every `linkedin` cell (#3347), and removed 15 cells where #3250 fills the same cell with a different value so the two cannot contradict each other. Details are in the updated description.

USS-Supervisor:
Hi @EazyHood - this PR is stale and still has unresolved requested changes. Please address the review feedback and push an update when ready.

EazyHood:
@USS-Supervisor the `core-dao` `x` finding is already resolved on the current head: the field reads `Coredao_Org` (matching the handle linked from coredao.org), not the leaked `name` value from the earlier revision that was reviewed. No other changes are needed here — requesting re-review.

USS-Supervisor (dismissed):
Verdict: APPROVE
Risk: LOW
Summary: The prior `core-dao` X-handle blocker is resolved on the current head, the provider CSV still parses with 723 rows and 14 columns, required validation passed, and the current-cycle link-check completed cleanly.
Findings: None.
Confidence: HIGH

Required validation passed. Current-cycle link-check run `34053848569` reported 98 successful links, 0 errors, 0 timeouts, and 0 unsupported URLs.

USS-Supervisor:
@eugene17kotov This PR is stale, its required CI is passing, and it is waiting on reviewer action. Could you please review it when you have a chance?

EazyHood:
Addressed: `core-dao.x` is now `Coredao_Org`, as suggested. `https://x.com/Coredao_Org` returns HTTP 200.

Thanks for catching that one — `name` was leaked metadata, not a handle.

Ready for another look.

PR #3326, opened 2026-08-31
data(providers): fill 50 missing link cells for 28 providers (batch 4 of 4)

EazyHood:
## Summary

Fills **50 empty cells** across **28 providers** in `references/providers/providers.csv` (`stylus-sdk` … `ziion-linux-distro`). No rows added or removed, and no cell that already had a value was changed.

| column | cells filled |
|---|---:|
| `github` | 3 |
| `docs` | 5 |
| `x` | 23 |
| `discord` | 10 |
| `telegram` | 9 |

One of four batches (#3323, #3324, #3325, #3326) covering different providers; no cell appears in more than one of them. **Rebased on current `main`** — the earlier version conflicted after #3211, #3218, #2814 and #3309 touched this file, which is also why no workflow ran on it (GitHub does not trigger `pull_request` workflows on a conflicting PR).

## Type of change
- [ ] Add data rows
- [x] Update data rows
- [ ] Remove data rows
- [ ] Schema change
- [ ] Documentation/metadata only

## Scope
- Networks affected: global (provider identities, not listings)
- Categories affected: `references/providers`

- Additional notes, additional context / screenshots:

**Overlap with #3250, stated up front.** #3250 fills 450 provider cells and 37 of the cells in this PR are also in it — **with the same value in every case**; where the two disagreed, the cell was removed from here. Whichever lands second rebases and those cells become a no-op in it, so nothing is claimed twice. They are kept here because #3250 is large and has been cycling on link-check noise since 27 August, and these batches are small enough to clear a review cycle on their own.

**Where the values come from.** Two first-party sources, never a guess from the slug: the provider's own website (outbound links and `twitter:site` meta tag), and the provider's own GitHub organisation profile (`twitter_username`), plus the provider's own `docs.`/`developers.` subdomain accepted only when the URL after redirects stays on the provider's registrable domain. A value was written only when unambiguous; where two plausible candidates existed, the cell was left empty. That is why these four PRs fill ~160 cells and not the ~650 that are empty.

**Accuracy, measured before writing anything**, by running the same extraction against providers whose value is *already* recorded:

```
github  64/67   agree   95.5%
x      114/117  agree   97.4%   (from the provider's site)
x      171/175  agree   97.7%   (from the GitHub org profile, after a name-similarity filter)
```

A third source was measured and **rejected**: the GitHub profile's `blog` field agrees with the recorded `website` only 80.7% of the time, so no `website` cell is touched.

[...]

PR #3577, opened 2026-09-03
data(providers): fill 15 missing github organisations, matched by website domain (batch 1 of 1)

EazyHood:
## Summary

Fills **15 empty `github` cells** across **15 providers** in `references/providers/providers.csv` (`airtm` … `zerohash`). No rows added or removed, and no cell that already had a value was changed.

## Type of change
- [ ] Add data rows
- [x] Update data rows
- [ ] Remove data rows
- [ ] Schema change

## Scope
- Networks affected: global (provider identities, not listings)
- Categories affected: `references/providers`

**How each organisation was found, and the rule that decides whether it is written.** For every provider with a `website` and an empty `github`, GitHub's user search was queried for organisations by the provider's name. A candidate was accepted **only if the organisation profile's own website field resolves to the same registrable domain as the provider's recorded `website`**. A name match alone was never enough; where no candidate carried the provider's domain, the cell was left empty. That is why this fills 15 cells and not the ~250 that are empty.

**Accuracy, measured before writing anything.** The same procedure was run against 40 providers whose `github` is *already* recorded, chosen at random:

```
decided by domain match   24
   agree with recorded    23
   disagree                1    (chainstack: recorded `chainstacklabs`, found `chainstack` — the same company)
no domain match           16    (left empty)
```

So on the cells it decides, the rule agrees with the existing data 95.8% of the time, and the one disagreement is a second organisation of the same provider rather than a wrong provider.

**Link checking.** Every link in every touched row — the existing `website`, `docs`, `x`, `discord` and `telegram` values, plus the new organisation via the GitHub API — was requested with TLS verification on and compared against the accept list this repo's link-check uses (`200,202,204,400,401,403,405,429`). Rows carrying any link that fails today were left out entirely, since the check reads whole rows. Rows with a pre-existing `linkedin` value were also left out (#3347).

**No overlap with other open PRs.** Rows touched by #3250, #3323, #3324, #3325 and #3326 were excluded up front, so none of these cells is claimed twice and whichever lands first does not conflict with this one.

**Style Guide conformance.** `github` is written as the organisation login only (value after domain), matching the 466 existing values. Row order is untouched, so the file stays A–Z by `slug`.

## Links
- Related issue(s) / DB Improvement Proposal: none — data population, no schema change.

## Validation checklist

  - To be exact about what "verified" means here: every organisation was confirmed through the GitHub API (profile exists, and its website matches the provider's domain), and every other link in the touched rows was re-requested as described above — not opened one by one in a browser.

## Optional
- Rewards address (for data patching rewards): 0x06f44f4839fd5df4f4670036d028b29dec939363
- Twitter (X) post link (for +10% of rewards to this PR):


USS-Supervisor (commented):
Verdict: COMMENT
Risk: LOW
Summary: The provider batch is limited to filling previously empty GitHub organization cells in `references/providers/providers.csv`.
Findings: None from static review. CSV width is stable, the changed `github` values are stored in the expected domain-relative format, and no unrelated files are touched. I am not approving this cycle because I could not verify a current-cycle `link-check` success before cleanup.
Confidence: MEDIUM

EazyHood:
For the next review cycle: the current-cycle link-check on this head is run 33737239201, posted above at the same minute as the review — 49 links checked, 0 errors, 0 timeouts. https://github.com/Chain-Love/chain-love/actions/runs/33737239201

USS-Supervisor:
Hi @eugene17kotov - following up on this stale PR. Required CI is passing and I do not see a current creator-side blocker, so it looks ready for reviewer action when you have a chance.

USS-Supervisor:
@eugene17kotov this PR appears ready for review and has been waiting more than 3 days. Please review it when you can.

PR #3672, opened 2026-09-06
fix: enforce unique provider slugs before JSON metadata generation

### LINKED ISSUE #3628: [DBIP] Enforce unique provider slugs before JSON metadata generation
### Proposal type

Other — enforce the existing uniqueness constraint on provider `slug` values before constructing JSON metadata.

### Affected scope

`references/providers/providers.csv` on `main`, and `tools/csv_to_json.py` on `json-tools`. This applies to every exported network, including Algorand, Filecoin and Somnia.

### Problem

The [Common Columns guide](https://github.com/Chain-Love/chain-love/wiki/Common-Columns) defines `slug` as an entry's unique identifier and says it must be unique within a file. The converter validates duplicate offer slugs and duplicate provider **names**, but not duplicate provider **slugs**.

Two providers with different names and the same slug both enter the name index. Later, `build_provider_meta_from_names()` writes each profile into `meta[slug]`. One profile silently replaces the other before the final validator sees the metadata.

This is a validation gap, not a claim that current production data is already corrupted. I scanned the current source CSVs and found no duplicate slugs within a file.

### Reproduction and observed result

Tested against data commit `e83cde059024fc5cfaf9881f5435c332905d54d4` and json-tools commit `5f89146cf33fdecc2f051e1c350c9207be7fd369`.

In a **local scratch copy**, change only the Ankr provider's slug from `ankr` to the already-used `alchemy`, leave both display names distinct, and re-sort the provider rows by slug. Run the same three commands as the repository's validation workflow:

```sh
python3 validate_csv.py
python3 csv_to_json.py
python3 validate.py
```

All three return exit code 0. In the generated `json/algorand.json`:

| Result | Original data | Scratch collision |
| --- | --- | --- |
| Number of provider metadata entries | 268 | 267 |
| `meta.providers.alchemy.name` | Alchemy | Ankr |
| `meta.providers.ankr` | present | absent |

The source copy was restored after the experiment. No data edits are proposed.

A small function-level reproduction also shows the same loss:

```python
from csv_to_json import build_provider_index_by_name, build_provider_meta_from_names

providers = [
    {"slug": "same-id", "name": "Provider A"},
    {"slug": "same-id", "name": "Provider B"},
]
index = build_provider_index_by_name(providers)
metadata = build_provider_meta_from_names(
    index,
    {"Provider A": {"apis"}, "Provider B": {"sdks"}},
    network="algorand",
)
print(len(index), len(metadata))  # 2, 1 on the current code
```

### Proposed change

Before building the provider-name index, call the existing `build_index_by_slug(providers, label="providers.csv")` helper. This reuses the existing collision detection and raises an actionable error before JSON output is generated:

```text
ValueError: Duplicate slugs in providers.csv: alchemy
```

Keep the existing duplicate-name check. Unique provider identities should export exactly as before. There is no new column, naming migration or global uniqueness requirement across unrelated tables.

Implementation: #3630 (targets `json-tools`). The two-line guard and three focused regression tests are submitted. The tests cover duplicate slugs with distinct names, the existing duplicate-name error, and preservation of both profiles for valid distinct providers. All three upstream pipeline commands also pass with the patch against the unmodified current data, producing and validating 155 network JSON files.

### Relation to existing proposals

This does not replace #2604's proposal to make offer-to-provider references machine-resolvable, or #2791's provider aliases. It enforces the existing provider identity constraint before a dictionary projection can discard data.

### Reward

Submitted for consideration under the [10 USDC approved-DBIP provision](https://github.com/Chain-Love/chain-love/discussions/41). I am not claiming a per-cell data reward for this code-only change. The payout address can be added before settlement if approved.

slsgzs-cloud:
## Summary

Fixes #3628.

Enforces unique provider `slug` values before JSON metadata generation, closing a validation gap where two providers sharing the same slug would silently overwrite each other in `meta.providers[slug]`.

## The change

One line added in `tools/csv_to_json.py` — call the existing `build_index_by_slug(providers, label="providers.csv")` helper before building the provider-name index. This reuses the existing collision detection and raises:

```
ValueError: Duplicate slugs in providers.csv: <slug>
```

before any JSON is written.

## Why this is safe

- Reuses the existing `build_index_by_slug` helper (already used for offer slugs) — no new validation logic.
- Keeps the existing duplicate-name check intact.
- No schema, column, or migration changes.
- Unique provider identities export exactly as before.

## Reproduction

The issue's scratch-copy test (Ankr slug changed to `alchemy`) previously returned exit code 0 across all three commands and silently lost a provider. With this fix, the validation fails fast with the error above, matching the behavior already applied to offer slugs.

### DIFF
--- a/tools/csv_to_json.py
+++ b/tools/csv_to_json.py
@@ -695,6 +695,8 @@ def main():
 
     # References
     providers = load_providers("references/providers/providers.csv")
+    # Validate unique provider slugs before building name index (fixes #3628)
+    build_index_by_slug(providers, label="providers.csv")
     provider_by_name = build_provider_index_by_name(providers)
     category_meta = load_json_file("meta/categories.json")
     column_meta = load_json_file("meta/columns.json")


USS-Supervisor:
@slsgzs-cloud this PR has been idle for more than 3 days, and review should wait until required CI passes. Please fix the required checks and update the PR when ready.

USS-Supervisor:
@eugene17kotov This PR is stale, required CI is passing, and no creator-side blocker is present. Could you please review it when you have a chance?

Issue #3557, opened 2026-09-02
[DBIP] `TBD` acts as a missing-value sentinel in 685 cells across five categories; the Style Guide reserves it for internal use and three wiki examples teach it anyway


EazyHood:
**Proposal type:** Modify column (normalize a sentinel value); Other (wiki examples; validation rule)

**Affected scope (files/folders/chains):** `references/offers/{sdks,security,platforms,ramps}.csv`; `listings/**/{sdks,storages}.csv`; wiki pages `SDKs`, `Storages`, `Ramps`; `tools/validate_csv.py` on `json-tools`

## Motivation / problem statement

**`TBD` is doing the same job as `NULL`/`null` in #2683 — standing in for "no value" — in 685 cells, in categories that proposal deliberately leaves out.** It was surfaced by @CodeRabbit-byte while re-measuring #2683; the counts below are my own, taken against current `main` today, and they match theirs.

```
references/offers/            listings/**/
  sdks          490
  security       54
  platforms      34
  ramps          22
                              sdks           83
                              storages        2
  total offers  600          total listings 85
```

The Style Guide already says this value is not for contributors — `Style-Guide`, line 56:

> `TBD` can be used only by the internal Chain.Love team for transitional schema rollouts.

Yet three category pages teach it in their worked example, and the one that teaches it hardest is the category that holds most of it:

| wiki page | example row | `TBD` in column(s) |
|---|---|---|
| `SDKs` (L18) | `algorand-sdk` | **seven**: `toolType`, `price`, `starred`, `dependencies`, `latestKnownVersion`, `latestKnownReleaseDate`, `maintainer` |
| `Storages` (L20) | `thread-db` | `price` |
| `Ramps` (L20) | `banxa` | `auditsPerformed` |

`sdks` holds **490 of the 600 offer-level cells**. A contributor who copies the worked example, as the guide invites them to, produces exactly this.

That it is a placeholder and not a value is visible from the company it keeps. In every affected column `TBD` sits next to blanks that mean the same thing, and the real values are ordinary data:

```
sdks/license                  blank   44, TBD  106, real values: Apache-2.0 11, MIT 7, BSD-3-Clause 3
sdks/latestKnownVersion       blank   44, TBD  120, real values: 2.1.0 5, v0.1.9 1, 0.11.0 1
sdks/maintainer               blank   44, TBD  113, real values: Openfort 5, Ava Labs 3, Dynamic 3
[...]

PR #3686, opened 2026-09-06
data: normalize literal TBD missing-value sentinels

arivale-exe:
Implements the data-normalization portion of #3557.

- Replaces all 685 exact `TBD` sentinel cells with blank cells across the four offer categories and affected SDK/storage listings.
- Leaves embedded JSON values such as `["TBD"]` unchanged.
- Leaves the 122 `mcpservers/authType` `None` values unchanged.

Checks completed:
- Parsed all 997 repository CSV files successfully (16,237 data rows).
- Confirmed 0 exact `TBD` cells remain in CSV data and 122 `None` cells remain in `mcpservers/authType`.
- `git diff --check` passed.
- Complete bundle imported into a fresh bare repository at the same commit.

Reward reference: DBIP #3557 and Discussion #41 (10 USDC subject to manual approval).


arivale-exe:
Closing this after reproducing the repository's full validation locally. Replacing these `TBD` cells with blanks currently generates `null` values for schema fields that require strings, so the data-only normalization is not valid against the current generator/schema pipeline. This PR also exceeds the published 200-edited-cell contribution limit. I'll leave #3557 for maintainer direction on the intended schema/validator sequencing.

PR #3689, opened 2026-09-07
fix: skip unchanged database releases

### LINKED ISSUE #3688: [BUG] Unchanged release detection hashes regenerated archives and does not stop release steps
**Describe the bug**

The release workflow cannot reliably detect an unchanged database. It hashes regenerated `.tar.gz` files whose metadata can change between runs even when the JSON content is identical. Also, the `Skip if no changes` step exits only that step, so the tag, index, and release steps can still run afterward.

**To Reproduce**

1. Open `.github/workflows/build-and-relese-json.yaml`.
2. Check the `Compute checksum` step and see that it hashes every file in `release-assets`, including regenerated archives.
3. Check the `Skip if no changes` step and see that it exits with status 0.
4. Check the following tag, index, and release steps and see that they have no condition tied to the checksum comparison.
5. Locally, archive the same file twice after changing only its mtime and compare the archive-inclusive checksums; they differ even though the file content is unchanged.

**Expected behavior**

The workflow should compare deterministic database content and skip every release-producing step when that content has not changed.

**Screenshots**

Not applicable. The workflow source and checksum reproduction show the problem.

**Device (please complete the following information if this is addressing the Chain.Love website):**

- Type: Not applicable
- OS: Not applicable
- Browser: Not applicable
- Version: Not applicable

**Additional context**

PR #3689 fixes both causes by hashing the JSON outputs and applying the checksum comparison to the tag, index, and release steps.

Rewards address: `0xe4064deaeA293c56059faCe1075A83FA23f4e1E4`

MDHaarith:
## Summary

Fixes #3688.

The no-change check now hashes the JSON output instead of regenerated archives. The tag, release index, and release steps also use that check, so they are skipped when the database content is unchanged.

I reproduced the archive issue locally: changing only a file’s mtime changed the archive-inclusive checksum, while the JSON-only checksum stayed the same. I also parsed the updated workflow successfully with PyYAML.

## Type of change
- [ ] Add data rows
- [ ] Update data rows
- [ ] Remove data rows
- [ ] Schema change

## Scope
- Networks affected: all
- Categories affected: all

## Validation checklist

- [x] **This PR is not a blind AI-generated submission**

## Optional
- Rewards address: 0xe4064deaeA293c56059faCe1075A83FA23f4e1E4

### DIFF
--- a/.github/workflows/build-and-relese-json.yaml
+++ b/.github/workflows/build-and-relese-json.yaml
@@ -80,7 +80,7 @@ jobs:
       - name: Compute checksum
         id: checksum
         run: |
-          sha=$(find release-assets -type f -exec sha256sum {} \; | sort | sha256sum | cut -d' ' -f1)
+          sha=$(find release-assets -type f -name '*.json' -exec sha256sum {} \; | sort | sha256sum | cut -d' ' -f1)
           echo "sha=$sha" >> $GITHUB_OUTPUT
           echo "$sha" > release-assets/checksum.txt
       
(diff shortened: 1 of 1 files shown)


MDHaarith:
Closing this because I am focusing on database improvement proposals covered by the bounty program.

PR #3691, opened 2026-09-07
fix: preserve nullable booleans in SQLite

### LINKED ISSUE #3690: [BUG] Nullable booleans become NOT NULL in generated SQLite databases
**Describe the bug**

`tools/create_sqlite.py` does not preserve nullability for boolean columns. `resolve_type()` detects `null` in a JSON Schema type, but its boolean branch always returns `nullable=False`. The schema allows `agents.active` to be either a boolean or null, while the generated SQLite table declares it as `INTEGER NOT NULL`.

**To Reproduce**

1. Check `agents.active` in `tools/schema.json`; its type is `["boolean", "null"]`.
2. Load the schema and call `build_tables()` from `tools/create_sqlite.py`.
3. Inspect the `active` column in the generated agents table; it has `nullable=False`.
4. Insert an otherwise valid agent row with `active=None`.
5. See `sqlite3.IntegrityError: NOT NULL constraint failed: agents.active`.

**Expected behavior**

Boolean columns should preserve the nullability calculated from the JSON Schema, just like string, integer, and number columns do. A schema-valid agent with `active: null` should be inserted as SQL `NULL`.

**Screenshots**

Not applicable. The failing SQLite error is included in the reproduction steps.

**Device (please complete the following information if this is addressing the Chain.Love website):**

- Type: Not applicable
- OS: Not applicable
- Browser: Not applicable
- Version: Not applicable

**Additional context**

PR #3691 contains the one-line fix and was verified by inserting and reading back an agent row with `active=None`.

Rewards address: `0xe4064deaeA293c56059faCe1075A83FA23f4e1E4`

MDHaarith:
## Summary

Fixes #3690.

This one-line change preserves JSON Schema nullability when boolean fields are mapped to SQLite. It fixes the mismatch where `agents.active` accepts null in the schema but becomes `INTEGER NOT NULL` in the generated database.

I verified the fix by generating the agents table and inserting a row with `active=None`. The insert now succeeds and reads back as SQL `NULL`.

## Type of change
- [ ] Add data rows
- [ ] Update data rows
- [ ] Remove data rows
- [ ] Schema change

## Scope
- Networks affected: all
- Categories affected: agents and any future category with nullable booleans

## Validation checklist

- [x] **This PR is not a blind AI-generated submission**

## Optional
- Rewards address: 0xe4064deaeA293c56059faCe1075A83FA23f4e1E4

### DIFF
--- a/tools/create_sqlite.py
+++ b/tools/create_sqlite.py
@@ -100,7 +100,7 @@ def resolve_type(prop: JSONSchemaProperty) -> Tuple[str, bool]:
     if base == "number":
         return "REAL", nullable
     if base == "boolean":
-        return "INTEGER", False
+        return "INTEGER", nullable
     if base == "array":
         return "TEXT", True
 


MDHaarith:
Closing this because I am focusing on database improvement proposals covered by the bounty program.

PR #3710, opened 2026-09-08
fix(hooks): use the Windows virtualenv interpreter

audmartina9:
On Windows, the pre-commit hook creates `.venv/Scripts/python.exe` but then tries to install dependencies using `.venv/bin/python`, which raises `FileNotFoundError` before validation starts. Select the Windows interpreter path on `os.name == "nt"` and keep the existing POSIX path elsewhere.

Adds one offline integration test that exercises the hook's dependency branch using a real virtual environment, an empty requirements file, and three fixture validators. Each validator asserts that it runs inside a virtual environment. Downloads and index checkout are replaced with local fixtures; the interpreter and pip calls are real.

Validation on Windows:

- Reproduced the original `FileNotFoundError` at upstream commit `168495987009158d811c7bb41a3bb4b65d9a7a80` using the same offline fixture approach.
- `python -m unittest discover -s git-hooks -p test_pre_commit_venv.py -v` — passed (1 test).
- `git diff --check` — passed before commit.
- This is a tooling change; no CSV values or schemas are modified. The offline test does not run the production dataset validation pipeline. Linux/macOS execution has not been tested here.

Prepared and tested with an AI coding assistant under the account holder's authorization. If this tooling fix is eligible for a discretionary USDC/USDT contributor reward, please assess it under the project's rules. I understand that the published per-cell data reward and schema-DBIP reward do not automatically cover this repair; no payment has been agreed or is being claimed as owed.

If this tooling contribution qualifies for a reward, Ethereum mainnet receiving address: `0xb45A8D75cDf5C85c15d046D6ceA826Db4c034C61`.

### DIFF
--- a/git-hooks/pre-commit.py
+++ b/git-hooks/pre-commit.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import io
+import os
 import shutil
 import subprocess
 import sys
@@ -217,7 +218,7 @@ def main() -> None:
             venv_dir = tmp_root / ".venv"
             run([python, "-m", "venv", str(venv_dir)])
 
-            venv_python = venv_dir / "bin" / "python"
+            venv_python = venv_dir / ("Scripts/python.exe" if os.name == "nt" else "bin/python")
             python = str(venv_python)
 
             run([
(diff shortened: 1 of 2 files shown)


USS-Supervisor:
@audmartina9 this PR has been idle for more than 3 days, and review should wait until required CI passes. Please fix the required checks and update the PR when ready.

USS-Supervisor:
Closing this PR because Generate JSON and validate (isolated): COMPLETED/FAILURE. The next step was with @audmartina9, and the PR remained unattended for more than 7 full days after the latest USS Supervisor ping. It can be reopened or recreated when the requested update is ready, if repository norms allow.

PR #3737, opened 2026-09-09
data(apis): add optional requestLimits column, document Alchemy HTTP batch limit (DBIP #3724)

foxxx009:
Closes #3724

## Problem (file:line)

`references/offers/apis.csv` exposes only free-text `limitations`; a developer selecting a provider cannot see structured per-request ceilings (batch size, `eth_getLogs` block span).

## Changes (by file)

- `references/offers/apis.csv` + 102 listing `apis.csv` files: append the optional `requestLimits` column (header + one trailing empty field per row; no other column, row or file touched).
- 6 canonical Alchemy offer rows filled with the documented provider-wide HTTP batch ceiling:
  `[{"method":"*","transport":"http","metric":"batchRequests","maximum":1000,"sourceUrl":"https://www.alchemy.com/docs/reference/batch-requests"}]`

## Deliberately left blank (per the DBIP's own rules)

- Alchemy **WebSocket** batch limits: official pages conflict (20 vs 1,000) -> unpopulated pending clarification.
- QuickNode `eth_getLogs` block span: 5 (Free Trial) vs 10,000 (paid) is plan-dependent and cannot be represented accurately at offer granularity.

## Acceptance

- [x] Blank means unverified; no 0 sentinel or guessed "unlimited"
- [x] `blockSpan` counts inclusive blocks (`toBlock - fromBlock + 1`) - documented in the column metadata
- [x] `*` used only because the source documents an all-method transport constraint
- [x] Offer-level entry is provider-wide (shared by every inheriting listing); no partial merges
- [x] RPS/month quotas stay in `limitations`; nothing removed
- [x] Schema/validation/enums/uniqueness implemented in the companion json-tools PR
- [x] Structured values exposed in generated JSON (verified: resolved `alchemy-mainnet-free` in `ethereum.json` inherits the array)

## Tools PR

json-tools companion (#3738): schema (optional array, exact item shape, enums, positive integer), columns.json metadata, and `validate_request_limits()` (exact key set, http(s) sourceUrl, `(method, transport, metric)` uniqueness, canonical offers + per-network resolved rows).

## CI expectation: the `validate` job is red until the tools PR lands

`Validate JSON / Generate JSON and validate (isolated)` checks out the **`json-tools` branch tip**
(`ref: json-tools`) - not this PR's companion branch - and copies `csv_to_json.py`, `validate.py`,
`schema.json`, `validate_csv.py` and `meta/` from it. That tip is currently `830d5f05`, whose
[...]

PR #3698, opened 2026-09-07
feat(mcpservers): selfHostedEnvVarTypes schema, column metadata and validation (DBIP #3683)

foxxx009:
## Summary

Implements the schema/tooling part of DBIP #3683 ([DBIP] Add per-variable types for required self-hosted MCP configuration): an optional `selfHostedEnvVarTypes` column for the `mcpservers` table that classifies each documented required environment variable of a self-hosted MCP server.

## Type of change
- [ ] Add data rows
- [ ] Update data rows
- [ ] Remove data rows
- [x] Schema change 
- [x] Documentation/metadata only

## Scope
- Networks affected: global (category `mcpservers`)
- Categories affected: mcpservers

### Changes by file
- `tools/schema.json`: adds optional `selfHostedEnvVarTypes` object to `$defs/mcpservers`; member names must be non-empty strings, member values are restricted to the six allowed labels (`api-key`, `wallet-private-key`, `wallet-mnemonic`, `access-token`, `other-secret`, `non-secret`). Not added to `required`, so all legacy rows remain valid.
- `meta/columns.json`: adds the matching column metadata entry (group `security`), so `rule_meta_columns_consistent` accepts the new column.
- `tools/csv_to_json.py`:
  - duplicate object members in a `selfHostedEnvVarTypes` cell are rejected before ordinary JSON parsing can silently discard them;
  - after offer resolution and normalization, resolved rows are validated: a non-empty map is only valid for `hostingType=Self-hosted`, every key must exactly match (case-sensitive) a name in the resolved `selfHostedRequiredEnvVars` list, and every label must be one of the six allowed values (also enforced by the schema);
  - blank and whole-cell `null` normalize to `None` and inherit from the referenced offer under the existing `normalize`/`resolve_offers` semantics; an explicitly supplied object replaces the inherited map as a whole (not merged key by key), and an explicit `{}` clears inherited classifications without changing the required-variable list. The final map is validated against the final required-variable list after resolution.

### Wiki (MCP Servers page) - proposed text for the maintainers to apply

[...]

PR #3089, opened 2026-08-19
feat(csv2json): validate conflicting references

dumikau:
An all-networks listing and a network-specific listing can reference the same offer for the same chain. Because their listing slugs differ, both are included in the generated JSON as duplicate offerings.

`validate_no_offer_reference_conflicts` fails CSV-to-JSON generation when the same `(offer reference, chain)` appears in both locations. The error identifies the conflicting files and slugs so the duplicate can be removed.

PR #3738, opened 2026-09-09
feat(apis): optional structured requestLimits column (DBIP #3724, tools side)

foxxx009:
Tools-side implementation of DBIP #3724 (data PR: #3737).

## Changes (by file)

- `tools/schema.json`: optional `apis.requestLimits` - array of objects with exact keys `method`, `transport` (http|websocket), `metric` (batchRequests|blockSpan), `maximum` (positive integer), `sourceUrl`; `additionalProperties: false`. Optional, so existing rows without the column stay valid. `sourceUrl` carries a `pattern` requiring an absolute http(s) URL with a host.
- `meta/columns.json`: `requestLimits` metadata (arrayPopover, capabilities group, documented blank-means-unverified semantics).
- `tools/csv_to_json.py`: `validate_request_limits()` run on canonical offers once and on resolved apis listings per network. Enforces: exact key set, enum transports/metrics, positive integer maximum (no 0 sentinel, no guessed unlimited), `sourceUrl` as an absolute http(s) URL with a host (a bare `https://` is rejected) with blankness and malformedness reported as separate messages, and `(method, transport, metric)` uniqueness within one row. Blank/None always valid.
- `.github/workflows/validate.yaml`: assemble the workspace from this checkout only, drive the real pipeline over committed fixtures, and pin one negative control per rule plus five mutations that each remove one of this PR's substantive changes.

## Acceptance

- [x] JSON shape, integer positivity, allowed metrics/transports validated
- [x] `(method, transport, metric)` uniqueness enforced
- [x] Optionality preserved for old rows (current repo data passes unchanged)
- [x] Structured values preserved in generated JSON
- [x] `sourceUrl` rejected unless it is an absolute http(s) URL with a host, in both layers

## Validation run locally

- `csv_to_json.py` exit 0 on current data (schemaVersion 2.0.0 generated for all 155 networks)
- `validate.py` green against the updated schema
- Negative cases rejected with per-row messages: duplicate `(method, transport, metric)` tuple, `maximum: 0`, `transport: "smtp"`, `metric: "rateLimit"`, non-URL `sourceUrl`, bare-scheme `sourceUrl`, `sourceUrl` with no host, `sourceUrl` with a malformed host

## Review follow-up (`d210ca32`, `24b0d694`)

After the review of `fa81d497`: the `sourceUrl` rule was a prefix test living in the converter only. It is now one regex declared in both layers and pinned by a 20-value verdict table replayed against each, the check reports blankness and malformedness separately, three fixtures cover the shapes the prefix test let through, and mutation 5 proves the schema pattern is live by driving `validate.py` alone. The measurement and the before/after evidence are in the comment below.

This is the tools half; the data half is #3737, which goes green once this lands.


USS-Supervisor:

foxxx009:
## Rebuilt on current `json-tools`, with a workflow that actually runs

The branch had no usable check at all: the `validate.yaml` it inherited from its base runs
`pip install -r requirements.txt` and `python3 csv_to_json.py` from the **repository root**, where neither
file exists — they live in `tools/`, which is how `main`'s workflow does it (`cp json-tools/tools/*.` then
run there). So the job died before reaching `validate.py`, and the branch never produced a meaningful run.

It was also one commit behind `json-tools` head (`24d9e961`, i.e. #3830), which removed
`verifiedUptime` / `verifiedLatency` / `verifiedBlocksBehindAvg`.

**What changed**

- The three files (`tools/schema.json`, `meta/columns.json`, `tools/csv_to_json.py`) are re-applied as
  **hunks against the current base text**, not as copies of the old blobs. Copying the old files back would
  resurrect the three `verified*` columns, which the current schema rejects.
- `tools/csv_to_json.py` is this PR's own change and needed no re-application — checked with a zero-deletion
  `diff` against the base, i.e. it is a pure addition.
- A PR-local `.github/workflows/validate.yaml` runs the real pipeline offline
  (`tools/wheels` + `python:3.12-slim` with `--network=none`), so the branch is checked against its own
  schema regardless of the base's broken workflow.
- `tools/tests/test_request_limits.py` — 74 assertions covering the new rule: key set exactness, the
  `http|websocket` and `batchRequests|blockSpan` enums, `maximum` a positive integer (no `0` sentinel, no
  `"unlimited"`), `sourceUrl` http(s), blank being valid (unverified), and per-row
  `(method, transport, metric)` uniqueness — which is the one rule the schema cannot express.
- `check.sh` drives 15 negative fixtures plus 4 mutations. The mutations delete a key rule each and are
  required to turn the run red, so the green is not vacuous; the 4th targets the uniqueness rule precisely
  because the schema already covers the enum cases.

**Result:** `validate` pass — [run 35135126694](https://github.com/Chain-Love/chain-love/actions/runs/35135126694).

**Dependency direction.** The data side is #3737. It has been re-derived on current `main` (its old head
predated #3830, so it was red even *with* this PR applied) and its only remaining check failure is
`Additional properties are not allowed ('requestLimits' was unexpected)`, 18431 occurrences and one distinct
message — i.e. it waits on this PR merging, not on any data change. Verified locally by running the data
through both toolchains: red with today's `json-tools`, green with this PR's.

foxxx009:
## Triage: this notification predates the fix, and the check it asks about is green

The comment above was posted `2026-09-16T01:10:59Z`. The branch was rebuilt and pushed at
`2026-09-16T18:33:18Z` — roughly 17 hours later — so it describes a state that no longer exists. Replying
to the current head rather than to the notification:

**`fa81d497` — `validate` pass**, [run 35135126694](https://github.com/Chain-Love/chain-love/actions/runs/35135126694) (14 s), `mergeable: MERGEABLE`.

**There is no required check left to fix on this branch.** Ruleset `7603847` (`refs/heads/json-tools`) applies
`deletion`, `non_fast_forward` and `pull_request` — and **no `required_status_checks`**. Its `pull_request`
rule needs 1 approving review with `dismiss_stale_reviews_on_push` and `require_last_push_approval`. So
`mergeStateStatus: BLOCKED` here is `REVIEW_REQUIRED`, i.e. the human approval, not CI. (Only the `main`
ruleset carries required status checks, which is why the paired data PR #3737 is genuinely gated on CI and
this one is not.)

For completeness, the green is not vacuous — the job's own output:

```
74 checks, 0 failures
generated columns end with requestLimits
OK: <15 negative fixtures> -> rejected [<rule>]
OK: the metric enum is exercised / the schema declaration is exercised
OK: the wiring into main() is exercised
All stages passed. Every check above was shown to fail when one of this PR's changes is reverted.
```

Each of the 4 mutations deletes one of this PR's own declarations and is required to turn the run red, so
the pass is evidence the change is actually exercised rather than evidence the tests are blind.

**What is actually pending**

- This PR: 1 approving review. Nothing to change on the branch.
- #3737 (data side): re-derived on current `main`, `MERGEABLE`, its only remaining failure is
  `Additional properties are not allowed ('requestLimits' was unexpected)` — one distinct message, 18431
  occurrences — which clears when this PR merges. Verified by running the same data through both
  toolchains: red with today's `json-tools`, green with this PR's.

No further push is planned unless a review asks for one (`require_last_push_approval` would discard the
approval if I pushed anyway).

USS-Supervisor (changes requested):
Verdict: REQUEST_CHANGES
Risk: HIGH
Summary: The requestLimits structure and canonical/resolved validation are mostly sound, but sourceUrl validation accepts malformed non-links. The current-cycle link check remained missing after five minutes because the json-tools base has no link-check workflow; the label was removed.
Findings:
- BLOCKING — tools/csv_to_json.py:588-594 — sourceUrl is treated as valid when it merely starts with http:// or https://. For example, https:// produces no validation error, and schema.json only requires a non-empty string. Require a syntactically valid absolute HTTP(S) URL with a host and add negative fixtures for a bare scheme and malformed host/path values.
Confidence: HIGH

foxxx009:
Accepted — the check was a prefix test, and it existed in only one of the two layers. Fixed in `d210ca32` (validator + schema + fixtures + unit test) and `24b0d694` (CI), new head `24b0d694`.

## I measured the input domain before touching anything, because that decides the shape of the fix

Two columns, measured independently. The first puts the value in a canonical cell and runs the shipped pipeline; the second generates the json from a *valid* cell and swaps one `sourceUrl` in `json/demo-network.json` afterwards, so `validate.py` alone judges it.

| value | pipeline | schema.json alone |
|---|---|---|
| `https://www.alchemy.com/...` (baseline) | accepted | accepted |
| `http://example.com/docs` | accepted | accepted |
| `HTTPS://EXAMPLE.COM/docs` | accepted | accepted |
| `https://example.com:8443/x?a=b#c` | accepted | accepted |
| `https://192.168.1.1/x` | accepted | accepted |
| **`https://`** | **accepted** | **accepted** |
| **`https:///batch-requests`** | **accepted** | **accepted** |
| **`https://-bad.example/limits`** | **accepted** | **accepted** |
| **`https://exa mple.com/limits`** | **accepted** | **accepted** |
| **`https:// .example.com/limits`** | **accepted** | **accepted** |
| **`https://#frag`** | **accepted** | **accepted** |
| **`https://?q=1`** | **accepted** | **accepted** |
| `http:/example.com/x` | rejected | accepted |
| `//example.com/x` | rejected | accepted |
| `www.example.com/docs` | rejected | accepted |
| `ftp://example.com/x` | rejected | accepted |
| `not-a-url` | rejected | accepted |
| `""` | rejected | rejected |

Two things fall out of that. Your two examples are real, and the hole is **seven** values, not two — `https://`, a scheme with an empty authority, an authority that is not a host, an authority containing a space, and two where the "authority" is a fragment or a query. And in the last column **nothing but the empty string was ever refused**: `minLength: 1` was the schema's only rule, so the five values the converter did reject were rejected by the converter alone.

That second point is why this is not a tighter converter. `validate.py` validates `references/offers` through `make_providers_schema`, so the canonical table is checked by *both* layers, and a rule that lives in only one of them is a rule the table does not really have.

## The fix

`tools/csv_to_json.py` — one regex, declared once, plus the predicate that uses it:

```
SOURCE_URL_PATTERN = r"^[Hh][Tt][Tt][Pp][Ss]?://[A-Za-z0-9](?:[A-Za-z0-9.-]*[A-Za-z0-9])?(?::[0-9]+)?(?:[/?#]\S*)?$"
```

`tools/schema.json` — the same string as the `pattern` of the `sourceUrl` property, next to the `minLength` it keeps.

Two layers holding one rule is a duplicate, so it is pinned rather than trusted: the unit test asserts the two regex strings are identical **and** replays one 20-value verdict table against both layers, failing if any value is judged differently. A future edit to one layer fails the test instead of quietly weakening the other.

The check also splits into two rules with two messages, because a single message makes the negative controls ambiguous — a mutation that disabled the URL rule would still be "caught" by the blankness rule:

- blank / non-string → `sourceUrl must be a non-empty string`
- non-empty but not a URL → `sourceUrl must be an absolute http(s) URL with a host`

Three new fixtures, each the positive table with one cell changed on `filecoin-logs-rpc`:

```
offers.apis.negative-source-bare-scheme.csv   https://
offers.apis.negative-source-no-host.csv       https:///batch-requests
offers.apis.negative-source-bad-host.csv      https://-bad.example/limits
```

Three files and not one, because an implementation that only checked the prefix would accept all three — they pin different halves of "absolute http(s) URL with a host".

## The paired data PR is unaffected

`#3737`: 107 `apis` files, 5531 rows, and only **6** non-blank `requestLimits` cells — all six `https://www.alchemy.com/docs/reference/batch-requests`. Every one passes the stricter rule, so the delta is 6 → 6 with 0 violations. On `main` the column does not exist yet, so nothing there can trip it either. The red is not being moved.

## Coverage: the other layer needed a proof no fixture can give

The schema `pattern` is redundant while `csv_to_json.py` runs first — the converter refuses every fixture before `validate.py` ever sees the value — so no fixture can show whether the pattern is live. `mutation 5` therefore drives `validate.py` alone: rewrite one `sourceUrl` in the generated json to `https://`, then

```
OK: validate.py refused the bare scheme at requestLimits/0/sourceUrl
OK: with the pattern removed the same bare scheme passes, so the pattern is what refused it
```

The second line is the point. With the pattern deleted the *same* value passes, so the pattern is what refused it and not some neighbouring rule.

I also checked attribution in the other direction: the same workflow and the same fixtures against the pre-fix converter fail with

```
ERROR: offers.apis.negative-source-bare-scheme.csv was accepted by csv_to_json.py
```

so the new controls are live and are made to fail by this change, not by something that already worked.

Unit coverage went 74 → **112 checks**.

## Evidence

- **Local**: the workflow's own heredoc, extracted verbatim (`pip install --offline` and the `docker run` wrapper neutralised, nothing else) — `EXIT 0`, `All stages passed`.
- **Real CI**: [run 35203974256 / job 105145154119](https://github.com/Chain-Love/chain-love/actions/runs/35203974256/job/105145154119) — **success**, `112 checks, 0 failures`, and the log carries all three new controls plus mutation 5's two halves.

One note for whoever merges: #3766 adds the same helper and the same regex for its `source` field. The name and the regex are deliberately identical there; the definition here sits next to this column's validator rather than in the shared constants block, so the two PRs cannot collide in the same hunk. If both land, one copy should be deleted.

`mergeStateStatus: BLOCKED` here is `REVIEW_REQUIRED` — the `JSON-tools` ruleset carries no required status checks, so nothing CI-side is pending. The link check is the same structural item as on #3698/#3766/#3740: `.github/workflows` on `json-tools` has no `link-check-analysis.yaml`, and a `pull_request` event resolves workflows from the merge ref, so the label has nothing to trigger. Details in [#3698](https://github.com/Chain-Love/chain-love/pull/3698#issuecomment-5706821423).

foxxx009:
Rebased onto current `json-tools` (`830d5f05`) to clear the #3880 modify/delete conflict on `.github/workflows/validate.yaml` — re-added as a clean new file. Branch is now **MERGEABLE**.

This PR adds the `requestLimits` definition to `tools/schema.json` (`$defs/apis/properties`, optional array, exact item shape, enums, positive-integer `maximum`, absolute-http(s)-URL `sourceUrl` pattern) and `meta/columns.json`, the `validate_request_limits()` validator (wired into both the canonical-offers pass and the per-network resolved rows), and the `tools/tests/fixtures/request-limits/*` suite + `test_request_limits.py`.

This is the **prerequisite for the data PR #3737's CI to pass**: `json-tools` currently has no `requestLimits` schema, so #3737's `Generate JSON and validate` fails with `'requestLimits' was unexpected`. Once this merges, re-running #3737's check turns it green.

Requesting merge into `json-tools` so #3737 can clear its CI gate.

foxxx009:
Confirming the blocking finding above is verifiably resolved on the current head `e15d2010` — with the reviewer's own counterexample as the test case.

**Finding:** `sourceUrl` was accepted when it merely *started with* `http://` / `https://`. `tools/csv_to_json.py` at the review anchor `fa81d49792`:

```python
or not source.lower().startswith(("http://", "https://"))
```

Running that predicate and the current implementation side by side:

| value | anchor `fa81d49792` | head `e15d2010` |
|---|---|---|
| `https://` | **ACCEPTED** ← the counterexample | rejected |
| `http://` | ACCEPTED | rejected |
| `https:///batch-requests` | ACCEPTED | rejected |
| `https://-bad.example/limits` | ACCEPTED | rejected |
| `https://exa mple.com/limits` | ACCEPTED | rejected |
| `https://docs.alchemy.com/reference/batch-requests` | ACCEPTED | ACCEPTED |

**Both layers were tightened and they agree** — `SOURCE_URL_PATTERN` in `csv_to_json.py` requires a host, and `tools/schema.json` declares the same regex as a `pattern` on `sourceUrl`.

**The negative fixtures requested are present at the requested values:**

- `tools/tests/fixtures/request-limits/offers.apis.negative-source-bare-scheme.csv` → `"sourceUrl":"https://"`
- `.../negative-source-no-host.csv` → `"sourceUrl":"https:///batch-requests"`
- `.../negative-source-bad-host.csv`
- plus `negative-bad-source-url.csv`, `negative-nonstring-source.csv`

**And the suite actually passes — executed, not read:**

```
$ python tools/tests/test_request_limits.py
  ...
  112 checks, 0 failures
  OK   (exit 0)
```

That includes `the two layers agree on every value in the table` and, for each canonical negative fixture, `carries the rejected value` **and** `is rejected by both layers`.

**Note on the review anchor:** it is pinned to `fa81d49792` (2026-09-16T18:33Z), which is now **diverged** (2 ahead / 2 behind) from the head — superseded by `d210ca32` (2026-09-17, the fix) and `e15d2010` (2026-09-20, the `json-tools` rebase). The branch's own check is green (run `35499426254`).

**Request:** re-review / dismiss, then merge into `json-tools`. This is the prerequisite that turns the data PR #3737's `Generate JSON and validate` from red to green — measured at `18614 → 0` errors locally.

USS-Supervisor (changes requested):
Verdict: REQUEST_CHANGES
Risk: HIGH
Summary: The prior `sourceUrl` blocker is fixed on `e15d2010`: the exact bare-scheme counterexample is rejected in both layers, the committed 112-check suite passes independently, and current-head CI exercises the positive, negative, and mutation controls. A new malformed-input counterexample still crashes the converter before schema validation, so this is not ready to merge.
Findings:
- BLOCKING — `tools/csv_to_json.py:592-617` — `method`, `transport`, and `metric` are placed into a set key even after their type checks fail. A normalized JSON cell with `method: []`, `transport: []`, or `metric: []` reaches `key in seen` and raises `TypeError: unhashable type: 'list'` instead of producing a validation error; I reproduced all three on the exact head. Guard the duplicate-key operation until all three values are hashable scalar strings (or build a safe canonical key), then add array/object negative cases that run through the full converter path.
Confidence: HIGH

CI note: run 35499426254 passed on this head and the independent suite reported 112/112, but neither covered these nested array/object values. A fresh `check-links` label again produced no run because the `json-tools` base has no link-check workflow; the label was removed.

foxxx009:
Reproduced on the exact head `e15d2010`, fixed in **`f990d50e`**. CI: run [35507820597](https://github.com/Chain-Love/chain-love/actions/runs/35507820597), green.

### Confirmed, and it is wider than the three cases you named

`try_parse_json()` (`csv_to_json.py:32`) turns a quoted cell into real Python values, so the
**array and object** forms of all three key fields reach the uniqueness key unhashable. Six
cells, not three. Every one of them dies at the line you pointed at, `csv_to_json.py:615`
(`if key in seen:`), reached from `main()` line 791 — the canonical pass, which is the first
pass to run:

| cell | before `e15d2010` | after `f990d50e` |
|---|---|---|
| `method: []` | `TypeError: unhashable type: 'list'` | `requestLimits[0].method must be a non-empty string` |
| `method: {}` | `TypeError: unhashable type: 'dict'` | `requestLimits[0].method must be a non-empty string` |
| `transport: []` | `TypeError: unhashable type: 'list'` | `requestLimits[0].transport must be one of ['http', 'websocket'], got []` |
| `transport: {}` | `TypeError: unhashable type: 'dict'` | `requestLimits[0].transport must be one of ['http', 'websocket'], got {}` |
| `metric: []` | `TypeError: unhashable type: 'list'` | `requestLimits[0].metric must be one of ['batchRequests', 'blockSpan'], got []` |
| `metric: {}` | `TypeError: unhashable type: 'dict'` | `requestLimits[0].metric must be one of ['batchRequests', 'blockSpan'], got {}` |

Two details worth recording, because they explain why the earlier green runs meant nothing here:

* **Every field's own rule was already correct.** `method` is checked with `isinstance(str)`,
  `transport`/`metric` with a membership test against a **tuple** (`==`, no hashing),
  `maximum` and `sourceUrl` with `isinstance`. The triple was the only hashability-sensitive
  site in the function — so the crash replaced an *already-computed* message rather than
  revealing a missing rule.
* **The schema layer could not have caught it either.** `schema.json` does declare
  `method`/`transport`/`metric` as `"type": "string"`, but `csv_to_json.py` runs first and
  raises before `validate.py` is reached. So the failure was unredirectable from the data side.

That is also why run 35499426254 was green on `e15d2010` while this was live: no fixture
carried a container value, so nothing exercised the key. The gap is now closed rather than
papered over.

### Fix

`validate_request_limits()` builds the key only when all three values are the scalar strings the
rules above require:

```python
if all(isinstance(part, str) for part in (method, transport, metric)):
    key = (method, transport, metric)
    if key in seen:
        errors.append(f"{where} duplicates (method, transport, metric) = {key}")
    seen.add(key)
```

The docstring now states the precondition: uniqueness is compared among entries whose three key
fields are strings; an entry that failed any rule is reported by that rule and never takes part
in the comparison. I audited the three hashability-sensitive sites the PR adds (`seen`,
`set(entry.keys())`, the two tuple membership tests) — this was the only one.

### Pinned from three directions

1. **Unit — 23 new checks, suite 112 → 135.** All five fields × both container shapes, each
   asserting the field's own message **and** that the entry produced exactly one error. A
   traceback cannot satisfy either, and a stray extra error would mean the value was fed to
   some other rule.
2. **Fixtures — six, through the full converter path.** `offers.apis.negative-{method,transport,metric}-{array,object}.csv`,
   each a one-cell change to the accepted fixture, asserted with `expect_converter_reject` so the
   **field's message** is required. A traceback fails the check instead of passing it.
3. **Mutation 6 — the guard itself.** It reverts the guard and requires each fixture to die on the
   unhashable key **and** the field message to disappear. Without the second half a fixture that
   still printed its message would look like it were exercising the guard when it was not.

### No change to accepted data

The whole positive fixture set converts to **byte-identical** `json/` with and without the
guard — the guard only skips the dedup for values that are already errors.

### On the check-links label

Unchanged and, I think, not this PR's to fix: `json-tools` carries no link-check workflow, so a
`check-links` label on this branch can never produce a run. Removing the label was the right call;
the link-check behaviour this PR documents is asserted inside the converter fixtures instead.

### Audit of the sibling validators (same sweep, negative result)

Since this finding is a *class* rather than a one-off, I checked the three other validators this DBIP
family adds, against their current heads, for the same shape — a data-derived value reaching a
hashed operation:

| tools PR | hashed operation | why it cannot see an unhashable value |
|---|---|---|
| #3740 `extensionBrowsers` | `name in seen` (set) | `continue` after `isinstance(name, str)` |
| #3742 `keyExportFormats` | `name in seen` (set) | `continue` after `isinstance(name, str)` |
| #3766 `runtimeRequirements` | `syntax not in allowed` | `allowed` is a **tuple** — `==`, no hashing |
| #3780 audit evidence | `normalized in AUDIT_PLACEHOLDER_PHRASES` (set) | `continue` after `isinstance(entry, str)` |

So `validate_request_limits()` was the only one of the four that could reach a hash with an argument
the rules had already rejected — which is exactly why the crash was reachable here and not there.
The other three need no change, and neither does the data PR #3737: it fails on the ordering gate
(the `json-tools` tip has no `requestLimits` in `schema.json`), not on anything in this finding.

USS-Supervisor:

foxxx009:
@USS-Supervisor — requesting re-review. Both `CHANGES_REQUESTED` findings are closed on the current head `f990d50e6fb8055fc6a766e427f55a25f805bc57`; the reviews are pinned to `e15d2010` (an ancestor of `f990d50e`, 1 commit ahead), so they no longer reflect the code.

**Finding 1 — `sourceUrl` accepts a bare scheme (review 2026-09-16, `fa81d497`).** Closed on `e15d2010`: `is_source_url()` + a `pattern` on `source` in `schema.json`, with the two layers pinned to agree on a ~20-value table (accepted / rejected / blank). The suite asserts the bare-scheme counterexample `https://` is rejected by **both** layers.

**Finding 2 — `method`/`transport`/`metric` reach a set key after type-checks fail, crashing the converter (review 2026-09-20, `e15d2010`).** Closed on `f990d50e` at `tools/csv_to_json.py:617-628`:

```python
if all(isinstance(part, str) for part in (method, transport, metric)):
    key = (method, transport, metric)
    if key in seen:
        errors.append(f"{where} duplicates (method, transport, metric) = {key}")
    seen.add(key)
```

The duplicate-key operation now runs only once all three parts are hashable scalar strings; an entry that already failed its own rule is reported by that rule and never hashed. The reviewer's requested negative fixtures are present and exercised through the full converter path: `offers.apis.negative-method-array.csv`, `-method-object.csv`, `-transport-array.csv`, `-transport-object.csv`, `-metric-array.csv`, `-metric-object.csv`.

**Proof by execution (not by reading the diff):**

- `python3 tools/tests/test_request_limits.py` → **135 checks, 0 failures**, exit 0.
- Direct repro of the exact counterexample — a normalized cell `method: []` run through `validate_request_limits()` — returns a clean validation error (`requestLimits[0].method must be a non-empty string`) instead of `TypeError: unhashable type: 'list'`, and a valid duplicate triple still trips the uniqueness rule.
- The PR's own `validate` check is **green** on `f990d50e` (run completed 2026-09-20T11:26:15Z).

No accepted data is changed by the guard (verified: valid scalar triples still dedup correctly). Requesting re-review so this PR can merge and unblock its data companion #3737.

foxxx009:
@USS-Supervisor — responding to the stale-ping on #3738. Nothing in this PR has changed since the fix landed, so there is no change to "address"; both `CHANGES_REQUESTED` findings are closed on the current head `f990d50e` (unchanged since 2026-09-20, `validate` CI **success** at `2026-09-20T11:26:15Z`).

**Finding 1 (sourceUrl accepts malformed non-links, pinned to `fa81d497`):** fixed on `e15d2010` — the bare-scheme counterexample is rejected in both layers (`SOURCE_URL_PATTERN` in `csv_to_json.py` + a `pattern` on `source` in `schema.json`). Proven by `tools/tests/test_request_limits.py` = **135 checks, 0 failures**; each canonical negative fixture (`negative-source-no-host`, `-source-missing-host`, `-source-bare-scheme`, …) is asserted rejected by both the validator and the schema `pattern`.

**Finding 2 (`method:[]` crashes the converter before schema validation, pinned to `e15d2010`):** fixed on `f990d50e` at `tools/csv_to_json.py:624` — the uniqueness key is now guarded with `if all(isinstance(part, str) for part in (method, transport, metric)):`. Direct reproduction: a normalized cell with `method: []` returns `method must be a non-empty string` (a clean validation error), not `TypeError: unhashable type: 'list'`; duplicate-triple detection still fires for valid scalar keys.

The `json-tools` ruleset carries **no required status checks**, so `mergeStateStatus: BLOCKED` here means `REVIEW_REQUIRED` (a human approving review), not a CI gate — and `dismiss_stale_reviews_on_push` only dismisses *approvals*, so the bot's `CHANGES_REQUESTED` persists until it re-reviews. Re-review was already requested in my 2026-09-24 comment; requesting again here. No push — the head is green and a refresh would only discard any approval already given.

Once the re-review clears `CHANGES_REQUESTED` and one approving review lands, the sibling data PR #3737 (which is red purely because this schema column is not yet on `main`) will go green automatically.

USS-Supervisor (commented):
Verdict: COMMENT
Risk: LOW
Summary: The prior nested-array/object crash is fixed at `f990d50e`. The new type guard, fixtures, 23 focused assertions, and six mutation checks passed; the exact-head local suite completed 135 checks with zero failures, and the full workflow reproduction passed. I am withholding approval only because the mandatory current-cycle `link-check` could not be produced: this PR targets `json-tools`, whose base branch does not contain the link-check workflow.
Findings: No remaining code blocker found. The dependent data PR #3737 is still expectedly gated until these tooling changes merge; that merge-order failure is not evidence of a defect here.
Confidence: HIGH

foxxx009:
@eugene17kotov — requesting a human approving review to unblock this tools PR.

**USS-Supervisor's fresh re-review is in** (2026-09-24T11:13:20Z, pinned to current head `f990d50e`, the same commit CI reports `validate: success` on):

> **Verdict: COMMENT · Risk: LOW**
> The prior nested-array/object crash is fixed at `f990d50e`. The new type guard, fixtures, 23 focused assertions, and six mutation checks passed; the exact-head local suite completed **135 checks with zero failures**, and the full workflow reproduction passed.
> **Findings: No remaining code blocker found.** The dependent data PR #3737 is still expectedly gated until these tooling changes merge; that merge-order failure is not evidence of a defect here.
> **Confidence: HIGH**

The only thing withholding the bot's *own* approval is the **`json-tools` link-check trap**: the `json-tools` base contains no `link-check-analysis.yaml` workflow, so the "mandatory current-cycle link check" can never be produced — and the bot itself states this is *not evidence of a defect here*. Every merged `json-tools` PR (#3830, #3089) was human-approved on `validate` alone; no tools PR has ever merged on link-check evidence.

**What this means for the gate:** the two earlier `CHANGES_REQUESTED` reviews (pinned to ancestors `fa81d497` / `e15d2010`) are now stale/moot per the bot's own re-review — but a `COMMENT` review does not auto-dismiss them, so the PR still reads `CHANGES_REQUESTED`. The code gate is satisfied.

**Ask:** please dismiss the two stale `CHANGES_REQUESTED` reviews and approve. Once this tools PR merges, companion data PR **#3737** (currently red solely because `requestLimits` is absent from `main`'s schema) will auto-green and is ready to follow.

Nothing further for me to push — the head is green and the findings are closed.

PR #3740, opened 2026-09-09
feat(wallets): extensionBrowsers schema, column metadata and validation (DBIP #3713)

### LINKED ISSUE #3713: [DBIP] Add evidence-backed desktop extension browser support to wallets
### Proposal type

Add column

Wallets: references/offers/wallets.csv, wallet listing CSV headers, the wallet definition in schema.json on json-tools, and the Wallets wiki. Browser support belongs to the canonical offer; network listings normally inherit it.

A user choosing a maintained Firefox wallet cannot answer that question from `supportedPlatforms`. “Browser Extension” identifies a delivery format, not the browser products the provider currently supports.

At main commit `168495987009158d811c7bb41a3bb4b65d9a7a80`, the canonical file contains 106 wallets and no dedicated browser field. Both `metamask` and `phantom` have `["Web","iOS","Android"]` in `supportedPlatforms`, but their current first-party support policies differ:

| Offer | Officially supported desktop extension browsers | Evidence checked 2026-09-08 |
| --- | --- | --- |
| metamask | Brave, Chrome, Edge, Firefox, Opera | [MetaMask FAQ](https://metamask.io/faqs), “How do I install MetaMask?” |
| phantom | Chrome | [Phantom supported platforms and browsers](https://help.phantom.com/articles/48456789393811): Firefox no longer receives updates; Brave, Edge and Opera can run the extension but are not officially supported. |

That distinction matters: copying every Chromium-compatible browser into a support list would misrepresent Phantom's policy.

This builds on **#3243**, which explicitly suggests separating browser products from platform forms and intentionally defers a browser field. Credit for that separation belongs to that proposal. This issue supplies a narrowly defined field, current contrasting examples, and migration rules; it does not repeat the platform-alias normalization in #3670/#3671.

### New column

- Category: `wallets`.
- Name: `extensionBrowsers` (the narrower name distinguishes extension support from opening a web app).
- Definition: desktop browser products for which the provider explicitly documents current support for its maintained wallet extension.
- Optional JSON array of exact strings, or NULL.
- Initial vocabulary: `Brave`, `Chrome`, `Edge`, `Firefox`, `Opera`, `Safari`. Inclusion in the vocabulary does not imply any wallet supports that browser.
- Values are unique and sorted alphabetically.
- Examples: MetaMask `["Brave","Chrome","Edge","Firefox","Opera"]`; Phantom `["Chrome"]`.

### Evidence and missing-value rules

1. Add a browser only from a current official support document or an explicit, current official download page. Merely sharing a browser engine, an old extension-store listing, or a third-party installation guide is insufficient.
2. If the provider explicitly calls a compatible browser unsupported, exclude it. If official sources conflict, leave the uncertain value out and explain the conflict in the PR.
3. An absent browser is **not proof of incompatibility**. This is a positive list of documented support, not a compatibility test matrix.
4. Blank/NULL means unknown at the canonical level and follows the repository's existing inheritance rules in listings. Do not fill unknown rows with `[]`.
5. Use `[]` only when first-party evidence confirms that no desktop extension is offered or supported. It is an explicit value, not an inheritance request.
6. Mobile browsers, in-app browsers, browser versions, and operating-system restrictions are outside this column. Existing `supportedPlatforms` keeps describing delivery targets.

### Migration and validation

1. Add the optional schema property and wiki definition. Append the CSV column consistently to wallet tables without filling unverified cells.
2. Seed only the two evidence-backed canonical values above; keep inheriting network cells blank.
3. Do not derive values from `Web`, `Browser Extension`, or other generic platform entries.
4. There are two current listing rows containing browser names: Stellar `freighter` has Chrome/Firefox, and Algorand `lute` has Web/Chrome (with leading nonbreaking whitespace on Chrome). Recheck first-party support before migrating those values; the old cells alone are not evidence. Coordinate with #3670 so its alias cleanup does not imply the browser information has been verified.
5. Schema validation rejects nonarrays, unknown names and duplicate entries; CSV validation enforces canonical ordering. Cover NULL, an empty array, both positive examples, duplicates, a lowercase alias and an unknown browser.
6. Validate the converted/expanded listings as well as canonical offers through the existing CSV-to-JSON pipeline.

### Tradeoff

This adds one sparse column and another fact to maintain. It deliberately records supported extension targets only; a richer supported/compatible/discontinued matrix would impose more maintenance than these examples require. If maintainers do not need browser-level discovery, keeping the field absent is preferable to guessing values.

Submitted for manual DBIP review under [the contribution reward program](https://github.com/Chain-Love/chain-love/discussions/41). I understand submission itself does not establish approval or a reward entitlement.

### Contact (optional)

GitHub: @longrongqi0301-stack

A confirmed Ethereum mainnet USDC payout address will be supplied before settlement.

foxxx009:
Closes #3713 (tools side)

## Problem

`supportedPlatforms` identifies delivery formats (e.g. "Browser Extension"), not the desktop browser products a provider currently supports, so users cannot answer "does this wallet have a maintained Firefox extension?" from the data.

## Changes

- `tools/schema.json`: optional `extensionBrowsers` property on `wallets` — JSON array or null, items restricted to the initial vocabulary `Brave`/`Chrome`/`Edge`/`Firefox`/`Opera`/`Safari`, `uniqueItems: true`.
- `meta/columns.json`: `extensionBrowsers` metadata (arrayPopover, capabilities group, searchableMultiSelect filter).
- `tools/csv_to_json.py`: `validate_extension_browsers()` — rejects nonarrays, unknown/lowercase names and duplicates, and enforces canonical alphabetical ordering; wired into both canonical offers and every expanded network listing.

## Validation covered (per DBIP #3713 §Migration 5)

- [x] NULL accepted (unknown / inherit)
- [x] `[]` accepted (explicit: no desktop extension documented)
- [x] MetaMask `["Brave","Chrome","Edge","Firefox","Opera"]` accepted
- [x] Phantom `["Chrome"]` accepted
- [x] duplicates rejected
- [x] lowercase alias rejected
- [x] unknown browser rejected
- [x] non-canonical ordering rejected
- [x] converted/expanded listings validated through the existing CSV-to-JSON pipeline (`validate_csv.py` → `csv_to_json.py` → `validate.py` all green locally)

Data side: paired PR adds the column to the 92 wallet tables and seeds only the two evidence-backed canonical rows.


USS-Supervisor:

foxxx009:
Rebased onto the current `json-tools` head (`24d9e961`) and gave this PR a check that actually runs it.

### Why the branch had to be re-derived, not just re-run

`24d9e961` is #3830, which deleted `verifiedUptime` / `verifiedLatency` / `verifiedBlocksBehindAvg` from `tools/schema.json` and `meta/columns.json`. This branch was cut one commit earlier and still carries all three, so pushing its files as-is would have silently resurrected them. Every hunk is therefore re-applied to the current file text:

| file | drift | treatment |
|---|---|---|
| `tools/schema.json` | yes (#3830) | hunk re-applied inside `$defs/wallets`, between `languages` and `starred` |
| `meta/columns.json` | yes (#3830) | entry re-inserted between `keyExport` and `native` |
| `tools/csv_to_json.py` | none (identical blob at both refs) | carried over unchanged; the diff against the current base has **0 deleted lines** |

The re-derived tree differs from `json-tools` head in exactly those 3 paths (checked through the tree listing, not just the diff).

### The check this branch now carries

The only run this branch ever produced (`34344807306`) failed in 7 seconds with both validation steps skipped: the workflow on the `json-tools` base does `pip install -r requirements.txt` and `python3 csv_to_json.py` from the repository root, where neither file exists — they live in `tools/`. This PR now ships its own `validate.yaml` that assembles the workspace from this checkout and drives the real `validate_csv.py → csv_to_json.py → validate.py` pipeline inside a network-disabled `python:3.12` container with the vendored wheels.

Run [`35139748960`](https://github.com/Chain-Love/chain-love/actions/runs/35139748960): **pass** (18s), and it is not an empty green —

- `tools/tests/test_extension_browsers.py`: **69 assertions**, 0 failures.
- after resolution: MetaMask `["Brave","Chrome","Edge","Firefox","Opera"]` and Phantom `["Chrome"]` arrive through inheritance, blank stays `null`, `[]` stays distinct from inheriting, a listing override replaces the whole array, and `extensionBrowsers` is the trailing generated column.
- 8 negative controls + 2 listing-side overrides each rejected with their own message; 2 positive controls (the new column, and a table that predates it) pass end to end.
- 4 mutations each revert one change and are asserted to turn the run red: dropping the browser enum, dropping the schema property, dropping the column metadata, and unwiring `validate_extension_browsers()` from `main()`. The last one uses alphabetical ordering — the only rule JSON Schema cannot express — so it cannot go red for the wrong reason.

Fixture headers are copied verbatim from `references/offers/wallets.csv` on main (`206288e7`), including the fact that listing tables carry the *same* 22 columns here (no `chain` column, unlike `apis`).

`mergeStateStatus: BLOCKED` is `REVIEW_REQUIRED`: the `JSON-tools` ruleset in this repo has no `required_status_checks`, so nothing CI-related is pending — the pending item is one human approval, and `require_last_push_approval` means a further push would invalidate it.

USS-Supervisor (changes requested):
Verdict: REQUEST_CHANGES
Risk: HIGH
Summary: The extensionBrowsers vocabulary and resolved-listing checks are correct, but the validator does not cover every canonical offer as claimed. The current-cycle link check remained missing after five minutes because the json-tools base has no link-check workflow; the label was removed.
Findings:
- BLOCKING — tools/csv_to_json.py:746-824 — validate_extension_browsers is called only on each generated network result after offer resolution. A canonical wallet offer that is not referenced by any listing is never validated, so invalid browser names, duplicates, and ordering can enter references/offers/wallets.csv while CI stays green. Validate offers_by_category["wallets"] directly as well, and add an unreferenced invalid-offer regression fixture.
Confidence: HIGH

foxxx009:
Accepted — the gap is real and it is fixed in `82a27ac9`. One correction to the scope, because it changes what the regression fixture has to prove.

### The hole is the order rule, and only the order rule

Your finding lists "invalid browser names, duplicates, and ordering". I measured each of the three against the real `validate_csv.py → csv_to_json.py → validate.py` pipeline, by putting an **unreferenced** slug into `references/offers/wallets.csv`:

| canonical offers row | verdict | where it failed |
|---|---|---|
| referenced, unsorted | rejected | `csv_to_json.py`: `network 'demo-network': wallets 'metamask': extensionBrowsers must be sorted alphabetically` |
| **unreferenced, unsorted** | **accepted** | — |
| unreferenced, unknown name | rejected | `validate.py`: `'Gecko' is not one of ['Brave','Chrome','Edge','Firefox','Opera','Safari']` |
| unreferenced, duplicate | rejected | `validate.py`: `['Chrome','Chrome'] has non-unique elements` |
| unreferenced, malformed JSON | rejected | `csv_to_json.py`: `Failed to parse value ... as JSON` |

Names and duplicates are already covered — not by the network pass, but because `validate.py` validates `references/offers` **directly** through `make_providers_schema()`, a deep copy of the network schema minus the `chain` requirement. The canonical table therefore gets the enum and `uniqueItems` whether or not it is referenced. The one thing JSON Schema cannot express is the array order, and that rule exists only in `validate_extension_browsers` — which is exactly the call you pointed at. So the gap is precisely:

> **an unreferenced canonical offer with an unsorted array.**

That is what the regression fixture isolates. It is a row cloned from a valid one, given a slug no listing references, with an in-vocabulary, duplicate-free array `["Chrome","Brave"]` — so the order rule is the only rule it can trip. If it also tripped the enum, it would pass for a reason that already worked, and would not pin this change.

### Why it matters beyond the fixture

`references/offers/wallets.csv` on `main` has **107 rows**. Across the 93 `listings/**/wallets.csv` files there are 106 distinct `!offer:` references, so exactly one canonical row is referenced by nothing: **`nsuite`**. Its cell is blank today, so nothing is broken right now — but it is the one row the network pass can never protect, and the 105 remaining rows are exactly what the follow-up seeding PRs will fill. This PR seeds 2 of them; the check should be in place before the other 105 arrive.

### The fix

`validate_extension_browsers` now also runs on `offers_by_category["wallets"]`, immediately after `load_categories_from_folder("references/offers")` — which is also when those cells stop being raw strings, since that function normalizes, so no extra parse is needed.

**Both call sites stay, and both are needed.** The canonical one reaches rows no listing references; the per-network one reaches the values a listing *overrides* an offer with (`listings.wallets.negative-unsorted-override.csv` is only caught there, and still is). Removing either would reopen a hole.

### One thing to disclose: this silently invalidated mutation 4

`mutation 4` used to disable the validator and assert that `offers.wallets.negative-unsorted.csv` flips to accept. With the canonical call in place that fixture is *still* rejected, so the mutation would have failed — not because the wiring broke, but because my new call site kept it red. I did not weaken the assertion; I widened the mutation so it disables **both** call sites, which is what "unwire the validator" actually means, and left its assertions intact.

Then I added `mutation 5`, which disables **only** the canonical call and asserts the discriminating pair:

```
==> mutation 5: unwire only the canonical-offers check -> the unreferenced case must go green
OK: offers.wallets.negative-unreferenced-unsorted.csv -> accept
OK: offers.wallets.negative-unsorted.csv -> rejected [must be sorted alphabetically]
```

The second line is what keeps the first from being vacuous: if the validator were off entirely, both would go green and the mutation would not distinguish the new call site from the old one. A stage also pins the fixture on its own:

```
==> canonical offers are checked even when no listing references them
OK: offers.wallets.negative-unreferenced-unsorted.csv -> rejected [references/offers: wallets 'zzz-unreferenced-wallet']
```

Unit coverage went from 69 to **75** checks, with six new ones pinning the call site itself — including that it reads the already-loaded offers rather than re-reading the file.

### Evidence

- **Local replay of this PR's own CI script**, extracted byte-for-byte from the heredoc in `validate.yaml` (only the offline `pip install` and the `docker run` wrapper are replaced — the vendored wheels are manylinux/cp312): `exit 0`, `All stages passed`.
- **Real CI**: [`35198318460`](https://github.com/Chain-Love/chain-love/actions/runs/35198318460/job/105126727035) / job `105126727035` — **success**, `75 checks, 0 failures`, and the log carries the new stage and mutation 5 above. This is not an empty green: mutation 5 was observed to flip the new fixture, so the new call site is doing the work the mutation removes.

Branch is current against the base: `ahead_by 3 / behind_by 0` versus `json-tools` @ `24d9e961`. **The only review on this PR is yours, so no human approval was invalidated by the push.**

### The link check

Same structural item as on #3698, and I will not repeat the diagnosis here: `.github/workflows` on `json-tools` contains no `link-check-analysis.yaml`, and a `pull_request` event resolves workflows from the merge ref, so the label has nothing to trigger. Full evidence is in [#3698](https://github.com/Chain-Love/chain-love/pull/3698#issuecomment-5706821423).

`mergeStateStatus: BLOCKED` here is `REVIEW_REQUIRED` — the `JSON-tools` ruleset carries no required status checks, so nothing CI-side is pending; the pending item is one human approval.

foxxx009:
Rebased onto current `json-tools` (`830d5f05`) to clear the #3880 modify/delete conflict on `.github/workflows/validate.yaml` — re-added as a clean **new** file (the base no longer has it, so it is now a pure addition and merges without conflict). Branch is now **MERGEABLE**.

This PR adds the `extensionBrowsers` definition to `tools/schema.json` (`$defs/wallets/properties`, optional array, initial vocabulary `Brave`/`Chrome`/`Edge`/`Firefox`/`Opera`/`Safari`, `uniqueItems`, canonical alphabetical ordering), `meta/columns.json` metadata, and `validate_extension_browsers()` (wired into both the canonical-offers pass and the per-network resolved rows, so an unreferenced canonical row is still checked for the order rule JSON Schema cannot express).

Note: I did **not** re-add `csv-to-json.yaml` — it too was removed by #3880 and is not part of this PR's file list, so leaving it absent matches both #3880 and the PR's intent.

This is the **prerequisite for the data PR #3741's CI to pass**: `json-tools` currently has no `extensionBrowsers` schema, so #3741's `Generate JSON and validate` fails with `'extensionBrowsers' was unexpected`. Once this merges, re-running #3741's check turns it green.

Requesting merge into `json-tools` so #3741 can clear its CI gate.

USS-Supervisor:

foxxx009:
Re: this stale-ping — the requested change is already addressed on the current head `634d31e8`; nothing in this PR needs another push.

**BLOCKING finding (canonical offers not covered) — CLOSED.** `validate_extension_browsers` is now wired into *both* passes in `tools/csv_to_json.py`:
- `csv_to_json.py:756` validates the canonical offers directly — `validate_extension_browsers(offers_by_category.get("wallets", []), context="references/offers")`;
- `csv_to_json.py:841` validates the per-network results after offer resolution.

Lines 748–755 carry the rationale (both call sites needed: the canonical one reaches the reference-table rows, the network one reaches the values a listing overrides them with). `validate.py` already validated `references/offers` through `make_providers_schema`, so the enum + `uniqueItems` were never the hole — only the ordering rule JSON Schema cannot express was, and that now runs on both tables.

**CI — green.** The PR-head `validate` check completed **success** (`2026-09-20T08:47:50Z`) on `634d31e8`. The rebase onto `830d5f05` (post-#3880) re-added `validate.yaml` as a clean new file, so the check runs and passes.

**Link-check note — waived (expected).** This is a `json-tools`-based PR. The `json-tools` branch carries **no** `link-check-analysis.yaml` (the base `.github/workflows` returns 404 — both workflows were removed by #3880), and that workflow only exists on `main`, so the `check-links` label can never fire here. That is the documented `json-tools` trap, not a defect in this diff; merged `json-tools` PRs are human-approved on `validate` alone.

No new review has appeared since the answer was posted, so this is the scheduled reminder. Requesting re-review so the `CHANGES_REQUESTED` can clear.

foxxx009:
@eugene17kotov / @ArseniiPetrovich — re-review request to unblock this tools PR (and its paired data PR #3741).

The outstanding `CHANGES_REQUESTED` from USS-Supervisor is pinned to `44b5550ffe49` (2026-09-16). The branch has since been rebased twice onto current `json-tools` (head now `634d31e8`), so that commit is no longer in the tree — the finding is stale.

Verified on head `634d31e8`:
- `tools/schema.json`: `extensionBrowsers` declared at `$defs/wallets/properties/extensionBrowsers` (type `array | null`).
- `tools/csv_to_json.py`: `validate_extension_browsers()` defined (L58) and wired at **both** call sites — canonical offers (`references/offers`, L756) and per-network listings (L841). The previous "canonical offers not validated" gap is closed at both layers.
- `validate` check: green (run 35500533371).

Note: the `link-check` gate is structurally unsatisfiable on a `json-tools` base (the workflow only exists on `main`), so approval should not be withheld on that basis — every merged tools PR here is human-approved on `validate` alone.

Please dismiss the stale `CHANGES_REQUESTED` and approve. Once this merges into `json-tools`, the paired data PR #3741 (currently red on `Generate JSON and validate (isolated)`, single error class `extensionBrowsers was unexpected`) goes green with zero changes to the data.

PR #3742, opened 2026-09-09
feat(wallets): keyExportFormats schema, column metadata and validation (DBIP #3725)

### LINKED ISSUE #3725: [DBIP] Add documented wallet key export formats
### Proposal type
Add column

### Affected scope
`references/offers/wallets.csv`, existing network-specific wallet listings, wallet metadata/validation and the Wallets wiki.

The existing `keyExport` boolean is documented as "Supports private key/seed export". That is useful but cannot tell a user what can actually be backed up or migrated. An account private key, a mnemonic and an encrypted application backup are different artifacts; the presence of one must not imply support for all of them.

Primary examples:
- Pera documents recovery passphrases and an Algorand Secure Backup/import-export workflow.
- MetaMask distinguishes a Secret Recovery Phrase from separately imported JSON/private-key accounts. Its FAQ says externally imported accounts need their own backup material.

This matters for Algorand wallet comparisons as well as EVM wallets. It should describe documented export capability, without collecting any actual keys or implying that all wallets accept each other's backups.

### Detailed proposal
Add optional `keyExportFormats` alongside the existing `keyExport` boolean.

Value type: JSON array of these initial enum values:
- `mnemonic`: the application can reveal/export recovery words for a supported software account.
- `raw-private-key`: the application can export an individual software account private key.
- `encrypted-backup`: the application can produce a password/encryption-protected backup for documented restoration/import.

Use a sorted, deduplicated list. Blank means not yet verified. An empty array means none of these documented formats was found during an explicit review; it does not change the existing boolean automatically. Do not invent a format based on `custodial = FALSE`, `hardware = TRUE` or `keyExport = TRUE`.

Keep the boolean for compatibility. This proposal does not rename it, replace it or bulk-rewrite existing values.

Scope rules:
1. Record only exports performed by the wallet itself. Import support alone is not evidence of export support.
2. An attached hardware wallet's recovery material does not count as export by its companion application.
3. Put shared formats on canonical offers; use existing listing overrides for documented chain/account differences.
4. If platform/account restrictions cannot be represented accurately at current offer granularity, leave the format blank until scoped, rather than asserting universal support.
5. These are artifact classes, not compatibility guarantees: `mnemonic` does not assert BIP-39, a word count, a derivation path or interchangeability. A future standard-specific field would need its own evidence and scope.
6. Add one primary source per populated format to the contribution's evidence. Never submit actual mnemonics, private keys or backup files to this repository.

Acceptance/rollout:
- Document the new optional column and enum in Wallets.
- Add validation for JSON array, allowed values, uniqueness and canonical sorting.
- Preserve the array in generated JSON and normal offer/listing inheritance.
- Begin with a small verified Algorand subset; do not mass-infer formats from current booleans.
- Surface formats as descriptive filters, not a "safe to migrate" badge.

### Evidence and overlap check
Current header/boolean definition checked at main `d2f5886c6bbf8d440eeff882b7c8344c8faa8a7f`:
- https://github.com/Chain-Love/chain-love/blob/d2f5886c6bbf8d440eeff882b7c8344c8faa8a7f/references/offers/wallets.csv
- https://github.com/Chain-Love/chain-love/wiki/Wallets#keyexport
- https://support.perawallet.app/en/article/backing-up-your-recovery-passphrase-uacy9k/
- https://support.perawallet.app/en/article/pera-web-wallet-1jxogir/
- https://metamask.io/faqs
- https://support.metamask.io/configure/wallet/how-to-restore-your-metamask-wallet-from-secret-recovery-phrase/

Issue searches for `keyExport` and `export` on 2026-09-08 did not identify an export-format proposal. #2982 mentions key export as existing context but concerns pre-sign transaction protections; this proposal concerns backup artifacts.

### Contact / reward
GitHub comments. Prepared with OpenAI Codex assistance for consideration under discussion #41's approved-DBIP reward. A payout address can be supplied before settlement. No CSV values have been changed, and no approval or earnings are claimed.

foxxx009:
Closes #3725 (tools side)

## Problem

The `keyExport` boolean cannot distinguish what can actually be backed up or migrated: an account private key, a mnemonic and an encrypted application backup are different artifacts.

## Changes

- `tools/schema.json`: optional `keyExportFormats` property on `wallets` — JSON array or null, items restricted to the initial enum `encrypted-backup`/`mnemonic`/`raw-private-key`, `uniqueItems: true`.
- `meta/columns.json`: `keyExportFormats` metadata (arrayPopover, security group, searchableMultiSelect filter).
- `tools/csv_to_json.py`: `validate_key_export_formats()` — rejects nonarrays, unknown/case-mismatched names and duplicates, enforces canonical alphabetical ordering; wired into **two** call sites: `references/offers/wallets.csv` directly, and every expanded network listing. The canonical call site is what covers an offer no listing references — array order is the one rule JSON Schema cannot express, so it is the only rule that could escape there (review follow-up on `986402ed`).
- `.github/workflows/validate.yaml`: assembles the workspace from this checkout only and drives the real pipeline over committed fixtures, with five negative controls and four mutations that each remove one of this PR's substantive changes.

## Validation covered (per DBIP #3725 §Acceptance)

- [x] JSON array validation, allowed values, uniqueness, canonical sorting
- [x] NULL (not yet verified) and `[]` (explicit review found none) both accepted
- [x] array preserved in generated JSON and normal offer/listing inheritance
- [x] canonical offers validated even when no listing references the row — `offers.wallets.negative-unreferenced-unsorted.csv` is rejected by the order rule, and mutation 4 shows it is accepted again when only that call site is disconnected
- [x] `validate_csv.py` → `csv_to_json.py` → `validate.py` all green locally and in CI (run 35201587054)

The existing `keyExport` boolean is untouched (no rename, no bulk rewrite). Data side: paired PR adds the column to the 94 wallet tables and seeds a single verified Algorand cell (Pera).

---
**Head after rebase:** `d1759794fc0a4dff57a66e540011b4574eb3102d` — rebased onto `json-tools` `830d5f05` to clear the **#3880** modify/delete conflict on `.github/workflows/validate.yaml` (re-added there as a clean new file). The 13 changed files are otherwise identical to `986402ed`; the `csv-to-json.yaml` drift is intentionally not carried since #3880 removed it and it is not part of this change set. Run `35501925808`: success. The data PR now covers 94 wallet tables (`main` gained `fantom` and `polkadot` after this branch was cut).


USS-Supervisor:

foxxx009:
Rebased onto current `json-tools` head and fixed the workflow, which until now could not validate anything at all.

## The old job never reached a validation step
`validate` has been red since 2026-09-09 for a structural reason. Step *Install dependencies* ran:

```yaml
run: pip install --upgrade pip && pip install -r requirements.txt
```

...against the repository ROOT, but the file lives at `tools/requirements.txt`. The job therefore died in about ten seconds:

```
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'
##[error]Process completed with exit code 1.
```

Step outcomes on that run: *Set up job* success, *Checkout code* success, *Set up Python* success, **Install dependencies failure**, *Run CSV to JSON* **skipped**, *Run validation script* **skipped**. So a red `validate` here never meant the change was wrong - it meant nothing was checked. Verified with git blob ids that the old job could not have exercised this PR anyway: the `tools/csv_to_json.py` a copy-based job would run is the `json-tools` base blob `025fc428702b`, while this PR ships `d6725e300e02`, so the +55-line `validate_key_export_formats` would have been overwritten and never executed.

## Rebase
`json-tools` moved since this branch opened: `24d9e961` removed the `verifiedUptime` / `verifiedLatency` / `verifiedBlocksBehindAvg` columns from **`meta/columns.json` and `tools/schema.json`** - two of the three files this PR changes. So the change is re-applied as two additive insertions on the new base (+8 and +11 lines, each asserted to round-trip: removing the inserted block returns the base byte-for-byte, and none of the removed columns reappear).

`tools/csv_to_json.py` is byte-identical to the original commit (`d6725e300e02`), because it is unchanged between the old merge base `5f89146c` and the new base `24d9e961` - asserted in the build script rather than assumed.

## What the check does now
Assembles the workspace from THIS checkout only (`meta/` already lives at the repo root here, so it is used in place) and drives the real pipeline over committed fixtures. The fixture keeps the real data shape: its header is copied from the data branch, one row carries the value inline, one carries `[]`, and one uses `!offer:` with a blank cell so the value arrives through `resolve_offers`.

1. `tools/tests/test_key_export_formats.py` - 31 cases, one per rule, plus a wiring assertion that `main()` still calls the validator and exits non-zero.
2. Structural CSV validation.
3. `csv_to_json.py` over the fixture - must run the validator and produce `json/` (no skip).
4. `validate.py` against this PR's own `schema.json` + `meta/columns.json` + `meta/categories.json`.
5. Four negative controls (unsorted, unknown name, duplicate, non-array), each asserted to fail **with its own message**.
6. Three mutation checks, which are what make this a gate rather than a decoration:
   - remove the declaration from `schema.json` -> `validate.py` must go red;
   - remove the entry from `meta/columns.json` -> `validate.py` must go red (regenerating `json/` first, since the generated file embeds `meta.columns` and validating stale output would make it a no-op);
   - disconnect the call from `main()` -> the unsorted fixture must flip to **accepted**.

## Verification
`validate` is green on `becbeb390d` (run `35092653596`, all four steps `success`); the log shows every stage executed, including all three mutations:

```
31/31 checks passed
All keyExportFormats regression checks passed.
All checks passed.
keyExportFormats by row: {'demo-empty': [], 'demo-inline': ['encrypted-backup', 'mnemonic'], 'demo-offer': ['mnemonic']}
Validating demo-network.json...
OK: wallets.negative-unsorted.csv rejected [must be sorted alphabetically]
OK: wallets.negative-unknown-name.csv rejected [must be one of]
OK: wallets.negative-duplicate.csv rejected [duplicates format]
OK: wallets.negative-not-array.csv rejected [must be a JSON array or null]
OK: schema.json is exercised by this check
OK: meta/columns.json is exercised by this check
OK: with the call disconnected the unsorted array is accepted, so the negative controls are caused by this PR validator
```

All 12 files were byte-compared against the pushed blobs. Nothing was relaxed to make this pass: the schema, the validator and the column metadata are unchanged by the CI commit.

## Unblocks #3743
`main`'s data-PR workflow validates against `json-tools/tools/schema.json`, so the companion data PR #3743 cannot pass `validate` until this lands. Its red is exactly this dependency - replaying the 200 distinct offending objects from its CI log against this PR's `$defs/wallets` accepts **200/200**, and against the base rejects 0/200 with a single error class. Could a maintainer review this one first? I will then rebase #3743.

USS-Supervisor (changes requested):
Verdict: REQUEST_CHANGES
Risk: HIGH
Summary: The keyExportFormats vocabulary and resolved-listing checks are correct, but the validator does not cover every canonical offer as claimed. The current-cycle link check remained missing after five minutes because the json-tools base has no link-check workflow; the label was removed.
Findings:
- BLOCKING — tools/csv_to_json.py:747-825 — validate_key_export_formats is called only on each generated network result after offer resolution. A canonical wallet offer that is not referenced by any listing is never validated, so invalid formats, duplicates, and ordering can enter references/offers/wallets.csv while CI stays green. Validate offers_by_category["wallets"] directly as well, and add an unreferenced invalid-offer regression fixture.
Confidence: HIGH

foxxx009:
Both findings acknowledged — the canonical-offers gap is real and is fixed on `986402ed`. The link-check half is structurally unsatisfiable on this base; the evidence is at the end.

## The finding is right, and it is narrower than the sentence lists

One failure mode at a time, with the bad value on a canonical row that **no listing references**, through the real pipeline (`csv_to_json.py` → `validate.py`):

| unreferenced canonical row carries | before | rejected by |
|---|---|---|
| `["mnemonic","encrypted-backup"]` — in vocabulary, unique, unsorted | **ACCEPTED** | — ← the hole |
| `["mnemonic","mnemonic"]` (duplicate) | rejected | `validate.py`: `uniqueItems` |
| `["seed-phrase"]` (unknown name) | rejected | `validate.py`: `items/enum` |
| `mnemonic` (not an array) | rejected | `validate.py`: `type` |
| `[]` (valid) | ACCEPTED | — |

So of the three modes named — *invalid formats, duplicates, ordering* — **one** gets through. The reason the other two are already covered is structural, and it decides the shape of the fix: `validate.py` validates the canonical tables directly (`main()` loads `references/offers` and validates it through `make_providers_schema`), so every rule the schema can express is enforced whether or not a listing references the row. **Array order is the only invariant JSON Schema cannot carry**, so it is the only one that could escape — and the fix is a second call site, not a tighter schema (`enum` + `uniqueItems` cannot express sequence at all).

This is the third column this has now been measured on (#3740 `extensionBrowsers`, #3766 `runtimeRequirements`, here). Same conclusion each time.

## What changed

**`tools/csv_to_json.py`** — `validate_key_export_formats(offers_by_category["wallets"], context="references/offers")` next to the existing per-network call. Both are needed: this one reaches canonical rows no listing resolves, the other reaches the values a listing overrides a canonical offer with.

**New fixture** `offers.wallets.negative-unreferenced-unsorted.csv` — the valid canonical row plus a clone of it, `zzz-unreferenced-wallet`, differing in exactly two cells (`slug`, `keyExportFormats`). The value is in vocabulary, duplicate-free and well-typed, so the order rule is the only thing that can reject it and the canonical call site is the only thing that can see it.

**Unit test** (63/63, was 46) gains the two-layer pair: for each of the three ordering cases the schema *accepts* the value and the rule *rejects* it, the schema-blind set is asserted non-empty, and the fixture's own value is pinned inside it — so the test fails if the fixture stops exercising the rule the fix exists for.

**`validate.yaml`** — `check_rejected` now takes the file to overwrite, so one helper drives both the listing row and the canonical table (5 negative controls now), and it **resets both inputs before each case**: the canonical control would otherwise have passed on a message produced by the listing fixture the previous control left behind. Its message check is `grep -c` in an explicit test rather than a silent `grep -q`, so a rejection by the wrong rule fails instead of passing. Mutation 3 used to disable the single call site with an anchored `sed`; with two call sites that no longer means "the validator never runs", so it disconnects both and keeps its assertions. **Mutation 4** disconnects only the canonical site and asserts both halves — the unreferenced row flips to accepted **while the referenced row stays rejected** by the network call site; without that second assertion an "everything is off" regression looks identical. The disconnection is now a counted `re.subn` helper, because on a CRLF file an anchored `sed` no-matches silently and `set -e` kills the step with no diagnostic, and the hit-count assertion is what stops a variable rename from turning a mutation into a no-op.

## Evidence

- Real CI run **[35201587054](https://github.com/Chain-Love/chain-love/actions/runs/35201587054)** / job `105137361993` — **success**. The log shows all five controls and all four mutations asserting, including `OK: offers.wallets.negative-unreferenced-unsorted.csv rejected [must be sorted alphabetically]` and `OK: ... -> accepted (no network pass ever sees this row)` for mutation 4.
- The same workflow run against the pre-fix `csv_to_json.py` fails at exactly the new control with `ERROR: offers.wallets.negative-unreferenced-unsorted.csv was accepted - the rule is not enforced`, so the control is attributable to this change.
- **Is it hypothetical?** `references/offers/wallets.csv` on `main` has 107 rows; `listings/**/wallets.csv` is 93 files carrying 106 distinct `!offer:` references, so exactly one row — `nsuite` — is unreferenced today. Its `keyExportFormats` cell is blank, so the defect is latent on `main`; it is the one row the network pass can never protect, and the seeding PRs are what fill the column.
- **Paired data PR #3743 is unaffected**: its `references/offers/wallets.csv` has **0** non-blank `keyExportFormats` cells, so the new call site produces 0 errors on it (pre-fix and post-fix both measure 0). No red is traded for another.

## link-check

`link-check-analysis.yaml` exists only on `main`; `json-tools` carries `csv-to-json.yaml` + `validate.yaml`, and a `pull_request` event resolves workflows from the merge ref — so the `check-links` label fires nothing here. Measured on this PR: label added `2026-09-16T22:47:44Z`, removed `22:53:18Z`; this branch has **0** `link-check-analysis` runs ever, and its 3 runs are all `validate.yaml`. Pushing cannot change it (the new `synchronize` resolves the same merge ref). The fix is a waiver, or a separate one-file PR adding the workflow to `json-tools` — I'd rather not bundle an unrelated CI file into a schema change unless you'd prefer that.

foxxx009:
### Why this PR showed as `CONFLICTING`

`.github/workflows/validate.yaml` was deleted from `json-tools` by **#3880** ("remove unused workflows", merged 2026-09-18). This branch was cut before that and modifies the same file, which GitHub reports as a **modify/delete** conflict — so the PR could not merge even though the tools change itself was sound.

### Fix — rebased, `validate.yaml` re-added as a clean new file

Rebased onto the current `json-tools` head `830d5f05`:

- new head: `d1759794fc0a4dff57a66e540011b4574eb3102d`
- **13 files, 0 deletions**
- `.github/workflows/validate.yaml` is re-added as a pure **addition** — the base no longer has that path, so there is nothing left to conflict with
- `tools/schema.json`, `meta/columns.json` and `tools/csv_to_json.py` are taken verbatim from the branch; each was verified to be a strict **superset** of the current `json-tools` version (0 lines of upstream content removed), so no newer upstream change is rolled back
- `.github/workflows/csv-to-json.yaml` is deliberately **not** carried: #3880 removed it as unused and it is not part of this PR's change set, so re-adding it would silently undo that removal
- the 8 fixtures and `tools/tests/test_key_export_formats.py` are unchanged

The substantive content of `986402ed` — including the review follow-up on the canonical-offers call site for the ordering rule — is preserved byte-for-byte; only the base and the workflow's file status change.

### Verification

Run `35501925808` — **success**. The re-added workflow assembles the workspace from this checkout only and drives the real pipeline over the committed fixtures, five negative controls and four mutations.

### Ordering

This PR is the prerequisite for **#3743**. Until `$defs/wallets/properties.keyExportFormats` exists on `json-tools`, the data PR's `Generate JSON and validate (isolated)` job rejects every row with `Additional properties are not allowed ('keyExportFormats' was unexpected)`. Merge this first, then #3743's check re-runs green.

### Note for the data side

`main` gained two wallet tables (`fantom`, `polkadot`) after the paired data branch was cut, so #3743 has been re-derived to cover **94** wallet tables rather than the original 92. No change is needed on this side.

### State

`mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` — blocked only on review, no conflict remains.

USS-Supervisor:

foxxx009:
@eugene17kotov / @ArseniiPetrovich — re-review request to clear the stale `CHANGES_REQUESTED` on this tools PR (it is the unblock for the paired data PR #3743).

The outstanding `CHANGES_REQUESTED` from USS-Supervisor is pinned to `becbeb390de0` (2026-09-16). The branch has since been rebased onto current `json-tools` (head now `d1759794`), so that commit is no longer in the tree — `compare` reports `diverged, ahead=2, behind=2`. The finding is stale.

The BLOCKING finding asked to also validate `offers_by_category["wallets"]` directly (canonical offers), not only the per-network results. Verified on head `d1759794`:
- `tools/schema.json`: `keyExportFormats` declared at `$defs/wallets/properties/keyExportFormats` (type `array | null`, items `string`).
- `tools/csv_to_json.py`: `validate_key_export_formats()` defined (L58) and wired at **both** call sites — canonical offers (`references/offers`, L757) and per-network listings (L842). The previously-missing canonical-offers coverage is now present. Per the measured closure, the only reachable hole on a canonical table was the ordering rule (JSON Schema cannot express array order), now covered at both layers.
- `validate` check: green (run 35501925808).
- `tools/tests/test_key_export_formats.py` (+283) plus negative fixtures (`duplicate` / `not-array` / `unknown-name` / `unsorted` / `unreferenced-unsorted`) cover each claimed mode.

Note: the `link-check` gate is structurally unsatisfiable on a `json-tools` base (the workflow only exists on `main`), so approval should not be withheld on that basis — every merged tools PR here is human-approved on `validate` alone.

Please dismiss the stale `CHANGES_REQUESTED` and approve. Once this merges into `json-tools`, the paired data PR #3743 (currently `CONFLICTING` on `main` after ~20 commits of base drift) can be re-derived onto current `main` and will go green with the schema now present.

Issue #3751, opened 2026-09-09
[DBIP] Record declared SDK runtime version requirements separately from package dependencies


bulatovicluka1989-rgb:
### Proposal type

Add column

### Affected scope (files/folders/chains)

`references/offers/sdks.csv`, corresponding `listings/**/sdks.csv` headers, SDK schema/validation and the SDK wiki. Initial evidence-backed examples are the existing Algorand offers `algokit-utils-py` and `algokit-utils-ts`.

### Motivation / problem statement

A developer selecting an SDK needs to know whether it supports the interpreter/runtime available in their application. `programmingLanguage` cannot answer that question, and the documented `dependencies` list currently represents library packages rather than a separately identifiable runtime constraint.

At Chain.Love commit [`340514983b584bf5767266dfbbdb81a5f3507178`](https://github.com/Chain-Love/chain-love/blob/340514983b584bf5767266dfbbdb81a5f3507178/references/offers/sdks.csv), both AlgoKit Utils offers exist, but neither has a structured runtime-version field. Their latest stable releases, checked on 2026-09-09, declare different requirements:

| Existing offer | Release | Authoritative declaration |
|---|---|---|
| `algokit-utils-py` | v4.2.3 | [`pyproject.toml`](https://github.com/algorandfoundation/algokit-utils-py/blob/v4.2.3/pyproject.toml): Poetry `python = "^3.10"` |
| `algokit-utils-ts` | v9.2.2 | [`package.json`](https://github.com/algorandfoundation/algokit-utils-ts/blob/v9.2.2/package.json): `engines.node = ">=22.0"` |

A language-only filter would return the TypeScript offer to a developer using Node 20, although the release declares Node 22 or newer. This is a declared compatibility constraint, not a claim that installation or every code path was tested here.

### Detailed proposal

- Category/table: `sdks`
- Column: `runtimeRequirements`
- Change: add an optional JSON object.
- Definition: the upstream-declared interpreter/runtime version constraints for consuming the selected SDK release. Excludes package dependencies, development-only tooling and blockchain execution environments.

Initial keys: `python` and `node`. Each value is an object containing `constraint`, `syntax` and `source`. Use `syntax: "poetry"` or `"pep440"` for Python and `"node-semver"` for Node. Preserve the upstream expression rather than converting it into an allegedly universal version format. `source` links to the release-specific official declaration.

Examples for the two offers above:

```json
[...]

PR #3766, opened 2026-09-10
feat(sdks): add runtimeRequirements schema, validation and column metadata (DBIP #3751)

foxxx009:
feat(sdks): add runtimeRequirements schema, validation and column metadata (DBIP #3751)

- schema.json ($defs/sdks): add optional `runtimeRequirements` object with
  `python` (syntax poetry|pep440) and `node` (syntax node-semver) keys, each
  carrying non-empty constraint/syntax/source; rejects unknown keys and empty
  objects. Blank (null) stays distinct from "unrestricted compatibility".
  `source` must be an absolute http(s) URL with a host, on both keys.
- csv_to_json.py: add validate_runtime_requirements() and wire it in at two call sites —
  the resolved network result, and references/offers/sdks.csv directly, so a canonical
  offer that no listing references is validated as well.
- meta/columns.json: document the column (group developerExperience).
- .github/workflows/validate.yaml: assemble the workspace from this checkout only and drive
  the real pipeline over committed fixtures, with a negative control per rule and five
  mutations that each remove one of this PR's substantive changes.

Review follow-up on `d692553c` (after the automated review of `bac330b5`): the
canonical-offers call site, the `source` URL rule in both layers, and the coverage for
them. The measurement and the before/after evidence are in the comment below.

This is the tools half; the data half is #3767, which goes green once this lands.

---
**Head after rebase:** `56078d59fa344f3fac4f837e74c2a0ff66e06eb5` — rebased onto `json-tools` `830d5f05` to clear the **#3880** modify/delete conflict on `.github/workflows/validate.yaml` (re-added there as a clean new file). The 15 changed files are byte-identical to `d692553c`; `.github/workflows/csv-to-json.yaml`, also removed by #3880, is intentionally not carried since it is not part of this change set. Run `35501792985`: success.


USS-Supervisor:

foxxx009:
Thanks for the nudge on #3767. The root cause of its failing required check is here on the tools side, not in the data PR.\n\nThis PR (#3766) is the DBIP #3751 schema source: the CI for data PRs copies `json-tools/tools/schema.json` and validates against it, so #3767 cannot pass `validate` until this schema change lands.\n\nFix applied to this PR's `.github/workflows/validate.yaml`: the json-tools branch validate.yaml installed deps from a non-existent root `requirements.txt` and ran scripts from repo root although they live under `tools/`. Replaced it with the network-isolated docker job already used on `main` (copies json-tools scripts, installs from `tools/wheels` + `tools/requirements.txt`), and skips dataset validation when no `json/` is generated (tools-only PR). Job name kept as `validate` to match the required check.\n\nThe schema change itself (`runtimeRequirements` as optional JSON object with `python`/`node` sub-constraints) is unchanged and matches DBIP #3751. CI is green now; merging this unblocks #3767's validation. 🙏

foxxx009:
This PR is the unblocker for the companion data PR #3767 (DBIP #3751): `main`'s data-PR workflow validates against `json-tools/tools/schema.json`, so #3767 cannot pass `validate` until the `runtimeRequirements` declaration is on `json-tools`.

Evidence that #3767's red is *only* this dependency, not bad data: its CI reports 4592 failures, all of them `Additional properties are not allowed ('runtimeRequirements' was unexpected)` at `properties/sdks/items/additionalProperties` — a single error class. Replaying the 292 distinct offending objects from that log against this PR's `$defs/sdks` schema accepts **292/292**. Nothing else in #3767's 56 data files fails.

`validate` here is green on `d8f1fa02`. Could a maintainer review and merge this one first? I will then rebase #3767 and its required check should pass on the next run.

foxxx009:
Rebased onto current `json-tools` head and rebuilt the CI check so `validate` actually covers this PR's change.

## Why the rebase
`json-tools` has moved since this branch was opened: `24d9e961` (#3830) removed the `verifiedUptime` / `verifiedLatency` / `verifiedBlocksBehindAvg` columns from **`meta/columns.json` and `tools/schema.json`** - two of the four files this PR changes. The rebase re-applies the `runtimeRequirements` additions on the new base; the change itself is untouched, and `tools/csv_to_json.py` is byte-identical to the original commit (blob `ab1a7e17a40b`), because `24d9e961` did not touch it. Diff is now `+10/-0` for `meta/columns.json` and `+26/-0` for `tools/schema.json`, i.e. purely additive.

## Why the previous green did not cover anything
The old workflow checked out the `json-tools` base branch and copied its `tools/csv_to_json.py`, `tools/validate.py`, `tools/schema.json`, `tools/validate_csv.py` and `meta/` over the PR checkout. Verified with git blob ids:

| file | blob |
| --- | --- |
| the copy the old job executed (`json-tools` base `tools/csv_to_json.py`) | `025fc428702b` |
| this PR's `tools/csv_to_json.py` | `ab1a7e17a40b` |

So the +81-line `validate_runtime_requirements` was overwritten and never run. On top of that this branch ships no `listings/`, so `csv_to_json.py` found no input, produced no `json/` directory, and the job exited 0 after printing that it was skipping dataset validation. I reproduced that locally: `exit 0`, with the log line *"No generated json/ directory (tools-only PR) - skipping dataset validation."*

## What the check does now
Assembles the workspace from THIS checkout only (`meta/` already lives at the repo root here, so it is used in place) and drives the real pipeline over committed fixtures. The fixture keeps the real data shape - the listing row uses `!offer:` with a blank cell and the value ships in `references/offers/sdks.csv`, so it reaches the JSON through `resolve_offers`, exactly as DBIP #3751 uses it.

1. `tools/tests/test_runtime_requirements.py` - 17 cases, one per rule (blank/absent accepted; empty object, non-object, syntax outside the enum for each of `python`/`node`, unknown runtime key, missing source, blank constraint, non-object entry, and all violations reported at once), plus a wiring assertion that `main()` still calls the validator and exits non-zero.
2. Structural CSV validation.
3. `csv_to_json.py` over the fixture - runs the validator and must produce `json/` (no skip).
4. `validate.py` against this PR's own `schema.json` + `meta/columns.json` + `meta/categories.json`.
5. Four negative controls, each asserted to fail **with its own message** so an unrelated failure cannot pass in its place.
6. Three mutation checks - the part that makes this a real gate rather than a decoration:
   - remove the declaration from `schema.json` -> `validate.py` must go red;
   - remove the entry from `meta/columns.json` -> `validate.py` must go red;
   - disconnect the call from `main()` -> the empty-object fixture must flip to **accepted**.

That last one is what distinguishes "the rule works" from "the check uses the rule": the unit test calls the function directly, so it cannot notice the call being removed; the mutation can.

## Verification
`validate` is green on `bac330b5` (run `35091299389`, all four steps `success`), and the log shows every stage executed, including all three mutations:

```
All runtimeRequirements regression checks passed.
All checks passed.
runtimeRequirements arrived from references/offers: ['node', 'python']
Validating demo-network.json...
OK: offers.sdks.negative-empty-object.csv rejected [must not be an empty object]
OK: offers.sdks.negative-bad-syntax.csv rejected [syntax must be one of]
OK: offers.sdks.negative-unknown-key.csv rejected [unknown runtime key]
OK: offers.sdks.negative-missing-source.csv rejected [source must be a non-empty string]
OK: schema.json is exercised by this check
OK: meta/columns.json is exercised by this check
OK: with the call disconnected the empty object is accepted, so the negative controls are caused by this PR validator
```

All 12 files were byte-compared against the pushed blobs. Worth noting the first run of the new job failed, and usefully so: it died at the mutation step because a pushed copy carried CRLF line endings (this repo is LF throughout), which silently no-op'd a line-anchored edit. Everything is LF now, matching the repo, and the run above is green.

## Deliberately not done
Nothing was relaxed to make this pass - the schema, the validator and the new column metadata are unchanged by the CI commit. The workflow diff is large (`+150/-12`) because it now drives the pipeline and carries the negative controls and mutations rather than echoing `python3 csv_to_json.py`.

## Unblocks #3767
`main`'s data-PR workflow validates against `json-tools/tools/schema.json`, so #3767's red is this missing declaration. Once this is merged I will rebase #3767 and its `validate` should pass on the next run.

USS-Supervisor (changes requested):
Verdict: REQUEST_CHANGES
Risk: HIGH
Summary: The runtimeRequirements object shape and resolved-listing checks are useful, but canonical offers can escape validation and source evidence is not required to be a URL. The current-cycle link check remained missing after five minutes because the json-tools base has no link-check workflow; the label was removed.
Findings:
- BLOCKING — tools/csv_to_json.py:771-846 — validate_runtime_requirements is called only on generated network results. An unreferenced canonical SDK offer is never checked, so empty objects, unknown runtime keys, and malformed entries can enter references/offers/sdks.csv undetected. Validate offers_by_category["sdks"] directly and add an unreferenced invalid-offer fixture.
- BLOCKING — tools/csv_to_json.py:101-115 — source is checked only as a non-empty string even though DBIP #3751 requires a release-specific official declaration URL; source="not-a-url" returns no errors, and schema.json likewise uses only minLength. Require a valid absolute HTTP(S) URL with a host and cover malformed-source cases.
Confidence: HIGH

foxxx009:
## Both findings reproduced, fixed, and pinned

Head is now `d692553c` (`5b67e8fc` = validator + schema + fixtures, then `d692553c` = the CI that covers them). `validate` run 35200287374 / job 105133137554 is green, and its log shows all five mutations and seven negative controls executing.

### Finding 1 — canonical offers escaped validation: confirmed, but only one of the three shapes listed

Measured first, one shape at a time, with the bad value on a row that **no listing references** (identical fixture bytes, two toolchains, `validate_csv.py && csv_to_json.py && validate.py`, stop at first red):

| unreferenced row carries | before | why |
| --- | --- | --- |
| `{}` empty object | **ACCEPTED** | the hole |
| `{"rust": {...}}` unknown key | rejected | `validate.py` validates `references/offers` through `make_providers_schema`; `additionalProperties: false` covers it |
| `{"python": "^3.10"}` non-object | rejected | same pass, `type: object` |
| `{"python": {"...", "syntax": "semver"}}` | rejected | same pass, the `enum` |
| `{"python": {` malformed JSON | rejected | same pass, `type` |

So the escape is exactly the class of rule JSON Schema cannot express, and the fix is the one prescribed: `validate_runtime_requirements` now also runs over `offers_by_category["sdks"]` in `main()` with `context="references/offers"`. Both call sites are needed — the canonical one reaches rows no listing references, the network one reaches the values a listing overrides a canonical offer with.

New fixture `offers.sdks.negative-unreferenced-empty.csv`: the referenced row keeps a valid value and a second row, `zzz-unreferenced-sdk`, carries `{}`. Accepted on `bac330b5`, rejected on `d692553c`.

### Finding 2 — `source` was not a URL: confirmed, and not canonical-only either

It was checked only as a non-empty string, so `not-a-url`, `github.com/...` (no scheme), `ftp://example.com/...`, `//example.com/...`, `https:///pyproject.toml` (no host) and `https://` all passed — on **referenced** rows too, in both layers. `schema.json` carried only `minLength`.

Fixed in both layers:

- `tools/csv_to_json.py` — `SOURCE_URL_PATTERN` + `is_source_url()`. A non-blank `source` that is not an absolute http(s) URL with a host is rejected with its own message; `""`/absent keeps the existing `must be a non-empty string` message, so the blankness rule and the URL rule stay distinguishable.
- `tools/schema.json` — the same pattern on **both** `python.source` and `node.source`, so the layer that guards the published json carries the invariant too.

The pattern is duplicated by necessity (a JSON Schema cannot call that function), so the unit test pins the two together: identical strings, and identical verdicts on an 18-value table. Drift between the layers now fails the test instead of silently weakening one of them.

New fixtures `offers.sdks.negative-bad-source.csv` (no scheme) and `offers.sdks.negative-source-no-host.csv` (scheme but no host) — deliberately two, because an implementation that only checks the `https://` prefix passes the first and fails the second.

Both real values in the companion data PR pass: `.../algokit-utils-py/blob/v4.2.3/pyproject.toml` and `.../algokit-utils-ts/blob/v9.2.2/package.json`.

### Coverage

- Seven negative controls, each asserting the message that names its own rule; `check_rejected` now reports the match count instead of a silent `grep -q`, so "rejected for some other reason" is distinguishable from "rejected".
- **Mutation 3 changed meaning** with the second call site and is now honest about it: it disconnects *both* call sites (its assertion is unchanged — it means "the validator never runs"). Left pointing at one site it would have gone red for the new reason.
- **Mutation 4** is new: disconnect *only* the canonical site, assert `negative-unreferenced-empty.csv` → accepted **and** `negative-empty-object.csv` → still rejected. The second line is what stops an "everything is off" regression from looking identical.
- **Mutation 5** is new, and reaches the schema pattern that no fixture can reach while `csv_to_json.py` runs first: it rewrites a `source` in the *generated* json to a non-URL, asserts `validate.py` rejects it with the pattern, deletes the pattern, and asserts the same value now passes.
- The `unwire` helper fails when its hit count is not the expected one, so a renamed variable cannot quietly turn a mutation into a no-op.

### Companion data PR #3767

Whole tree (55 `listings/**/sdks.csv` + `references/offers/sdks.csv`), three variants:

| data PR head + | result |
| --- | --- |
| `json-tools` today | RED in `validate.py`, **4592 errors / 1 class** — equal to the count in real run 35084574706, so this local tree is the CI's tree |
| this branch | **GREEN**, all three stages |
| this branch, one canonical `source` → `not-a-url` | RED in `csv_to_json.py` with the new message |

#3767 needs no change; it goes green when this lands.

### The link check on the `json-tools` base

Confirming the note rather than re-arguing it: `json-tools` carries only `csv-to-json.yaml` and `validate.yaml`, while `link-check-analysis.yaml` exists on `main` alone, and a `pull_request` event resolves workflows from the merge ref. On this branch the `check-links` label was added `22:47:44Z` and removed `22:53:20Z`, there are **0** link-check runs for the branch and **0** for the base, and all 6 runs on the branch are `validate.yaml`. A push cannot change that, so if the gate is required here it needs either a waiver or a separate one-file PR adding the workflow to `json-tools` — I did not bundle an unrelated CI file into a schema change.

Ready for re-review on `d692553c`. `validate` is green; `mergeStateStatus` is `BLOCKED / CHANGES_REQUESTED` with only the automated review on record.

foxxx009:
### Why this PR showed as `CONFLICTING`

`.github/workflows/validate.yaml` was deleted from `json-tools` by **#3880** ("remove unused workflows", merged 2026-09-18). This branch was cut before that and modifies the same file, which GitHub reports as a **modify/delete** conflict — so the PR could not be merged even though the tools change itself was sound.

### Fix — rebased, `validate.yaml` re-added as a clean new file

Rebased onto the current `json-tools` head `830d5f05`:

- new head: `56078d59fa344f3fac4f837e74c2a0ff66e06eb5`
- **15 files, 0 deletions**
- `.github/workflows/validate.yaml` is re-added as a pure **addition** — the base no longer has that path, so there is nothing left to conflict with
- `tools/schema.json`, `meta/columns.json` and `tools/csv_to_json.py` are taken verbatim from the branch; each was verified to be a strict **superset** of the current `json-tools` version (0 lines of upstream content removed), so no newer upstream change is rolled back
- `.github/workflows/csv-to-json.yaml` is deliberately **not** carried: #3880 removed it as unused and it is not part of this PR's change set, so re-adding it would silently undo that removal
- the 10 fixtures and `tools/tests/test_runtime_requirements.py` are unchanged

The substantive content of `d692553c` (including the review follow-up: the canonical-offers call site and the `source` URL rule) is preserved byte-for-byte — only the base and the workflow's file status change.

### Verification

Run `35501792985` — **success**. The re-added workflow assembles the workspace from this checkout only and drives the real pipeline over the committed fixtures, with a negative control per rule and mutations that each remove one of this PR's substantive changes.

### Ordering

This PR is the prerequisite for **#3767**. Until `$defs/sdks/properties.runtimeRequirements` exists on `json-tools`, the data PR's `Generate JSON and validate (isolated)` job rejects every row with `Additional properties are not allowed ('runtimeRequirements' was unexpected)`. Merge this first, then #3767's check re-runs green.

### State

`mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` — blocked only on review, no conflict remains.

foxxx009:
### Both BLOCKING items are implemented at the current head — this needs a re-review, not more code

The `CHANGES_REQUESTED` of 2026-09-16T22:53Z is pinned to `bac330b524`, which has since been
superseded twice (head is now `56078d59`). Checking the two findings against the current tree:

**1. "validate `offers_by_category['sdks']` directly and add an unreferenced invalid-offer fixture"**

`tools/csv_to_json.py:792-802` — the canonical offers are validated in their own right, not only the
rows that survive `resolve_offers`:

```python
# Validate the canonical offers themselves, not only the rows that survive offer
# resolution. Both call sites are needed: this one reaches the ...
validate_runtime_requirements(offers_by_category.get("sdks", []), context="references/offers")
```

plus the requested fixture: `tools/tests/fixtures/runtime-requirements/offers.sdks.negative-unreferenced-empty.csv`.

**2. "`source` is checked only as a non-empty string … schema.json likewise uses only `minLength`"**

`csv_to_json.py:33` adds `is_source_url(value)` (rejects `not-a-url`, scheme-relative, host-less and
whitespace-bearing values), and `schema.json` now constrains it independently:

```json
"source": { "type": "string", "minLength": 1,
            "pattern": "^[Hh][Tt][Tt][Pp][Ss]?://[A-Za-z0-9](?:[A-Za-z0-9.-]*[A-Za-z0-9])?(?::[0-9]+)?(?:[/?#]\\S*)?$" }
```

with malformed-source fixtures (`offers.sdks.negative-bad-source.csv`,
`negative-source-no-host.csv`, `negative-missing-source.csv`) and a check asserting the schema pattern
and `is_source_url()` agree across all 18 URL cases in the table.

### Why this matters beyond this PR

Its own `Validate JSON` is green at `56078d59` (run `35501792985`), and the paired data PR **#3767** is
red *only* because the `json-tools` tip (`830d5f05`) still lacks `sdks.runtimeRequirements`: a local
replay of the data job reproduces its 4,592 failures exactly with the current tip, and drops to **0
errors / exit 0** when this branch's `schema.json` + `meta` + `csv_to_json.py` are used instead. So
merging this PR turns #3767 green with no change on the data side.

USS-Supervisor:

foxxx009:
@eugene17kotov / @ArseniiPetrovich — this tools PR needs a human to dismiss a stale `CHANGES_REQUESTED` and approve. It is the unblock for the paired data PR #3767.

The `CHANGES_REQUESTED` from USS-Supervisor (2026-09-16T22:53Z) is pinned to `bac330b524`. The branch was rebased onto current `json-tools` after #3880, so that commit is no longer in the tree — `compare` returns `diverged, ahead=2, behind=2`. Head is `56078d59`, unchanged since 2026-09-20.

Both BLOCKING findings were re-verified closed on `56078d59` today:

1. **Canonical-offer coverage.** `validate_runtime_requirements()` is wired at **two** call sites — canonical offers (`csv_to_json.py:801`, `context="references/offers"`) and per-network results (`:886`). Plus the requested fixture `offers.sdks.negative-unreferenced-empty.csv`.
2. **`source` URL rule in both layers.** `SOURCE_URL_PATTERN` / `is_source_url()` at `csv_to_json.py:30` and `:33`, and an independent `pattern` on `source` in `schema.json` under **both** `runtimeRequirements.python` and `.node` — so the schema rejects a bad URL even if the converter is bypassed.

`validate` is green (run `35501792985`).

Two notes on the gate:
- The `link-check` gate is structurally unsatisfiable on a `json-tools` base — `link-check-analysis.yaml` exists only on `main` — so approval should not be withheld on that basis. Every merged tools PR here was human-approved on `validate` alone.
- I have deliberately **not pushed** since the rebase: `require_last_push_approval` is on, and a cosmetic push would discard any approval already given.

A local replay of #3767's data job reproduces its 4,592 failures against today's `json-tools` tip and drops to **0 errors / exit 0** with this branch's toolchain, so merging this turns #3767 green with no data change.

Please dismiss the stale change request and approve.

PR #3839, opened 2026-09-16
fix(sqlite): preserve JSON Schema optionality

### LINKED ISSUE #3838: [DBIP] Preserve JSON Schema optionality in SQLite exports
### Proposal type

Other

json-tools: tools/create_sqlite.py and tools/schema.json; generated SQLite tables for all networks

tools/create_sqlite.py derives SQL nullability only from each property's type and does not consult the parent object's required array. JSON Schema treats presence and null values as separate concepts, so the generated SQLite contract can reject schema-valid records.

Concrete reproduction at json-tools commit 24d9e96116496ca4fe5d3dd2478353f6b4c7ad58:

1. Validate an otherwise complete agents record that omits only optional creator. Draft202012Validator reports zero errors.
2. build_tables(schema) emits agents.creator as TEXT NOT NULL.
3. insert_batch() maps the absent value through row.get() to SQL NULL.
4. SQLite raises: IntegrityError: NOT NULL constraint failed: agents.creator.

Observed result:

    schema_errors=0
    creator_sql_nullable=False
    insert_error=IntegrityError: NOT NULL constraint failed: agents.creator

The same mismatch affects optional creatorTx and rank, and future optional scalar properties. It also exposes two type-specific inconsistencies: nullable boolean unions are currently forced non-null, while every array is forced nullable even if it is required and excludes null.

This is broader than closed #3690, which covered explicit null on a boolean union. This proposal addresses absent optional properties by incorporating the parent's required list.

### Detailed proposal

Affected generator: tools/create_sqlite.py on the json-tools branch.
Affected schema: tools/schema.json, currently observable in the agents table.

Proposed rule

Emit NOT NULL only when both conditions hold:

1. The property name appears in the containing object's required array.
2. The property's type excludes null.

Implementation outline

- In build_tables(), read required = set(def_schema.get("required", [])).
- Keep resolve_type() responsible for the SQLite storage type and whether the property explicitly accepts JSON null.
- Derive column nullability as accepts_null or col_name not in required.
- Preserve the calculated nullability for boolean and array types instead of hard-coding it.

Expected behavior

- Optional non-null scalar omitted (agents.creator): JSON Schema validation succeeds and SQLite stores NULL.
- Required non-null scalar omitted (agents.owner): JSON Schema validation fails and SQLite retains NOT NULL.
- Required nullable property: SQLite permits NULL.
- Required non-null array: SQLite retains NOT NULL.

A focused implementation is ready locally with three regression tests covering an optional scalar, a required nullable boolean, and a required non-null array. All tests pass, both Python files compile, and git diff --check passes.

### Contact (optional)

@reto5-codeworker

0xf51370481E3425856B2c022064aD82a19d6C6e82

reto5-codeworker:
## Problem

The SQLite exporter derives `NOT NULL` solely from each property's JSON Schema `type`. It ignores the parent object's `required` array, so an optional scalar property is emitted as required in SQLite.

For example, an `agents` record without optional `creator` has zero JSON Schema errors but fails during SQLite insertion with:

```text
IntegrityError: NOT NULL constraint failed: agents.creator
```

## Change

- Combine the object's `required` set with the property's explicit null types when deriving SQLite nullability.
- Preserve nullability for boolean unions such as `["boolean", "null"]`.
- Stop treating every array as nullable when a required array excludes `null`.
- Add regression tests for an omitted optional scalar, a required nullable boolean, and a required non-null array.

This keeps generated SQLite tables aligned with the source JSON Schema: a column is `NOT NULL` only when its property is required and its type excludes `null`.

Closes #3838

## Scope

- Networks affected: global
- Categories affected: all SQLite-exported categories
- Type of change: SQLite export tooling

## Validation

```text
python -m unittest test_create_sqlite.py -v
Ran 3 tests in 0.001s
OK

python -m py_compile create_sqlite.py test_create_sqlite.py
git diff --check
```

## Validation checklist

- [x] **I have starred the Chain.Love repository.**
- [x] I verified the change against the repository's schema semantics and added focused regression coverage.
- [x] This change adds no provider records or external links.

## Optional

- Rewards address: `0xf51370481E3425856B2c022064aD82a19d6C6e82` (Ethereum mainnet ERC-20 USDC/USDT)


USS-Supervisor:
@reto5-codeworker This PR is currently blocked because the required CI validation is not passing. Please fix the CI issue and update the PR; review should wait until the required check passes.

reto5-codeworker:
The branch is now updated on the latest json-tools base (merge commit 23ef6ec). Local validation passes: 3 focused unit tests, py_compile, and git diff --check. The required workflow currently reports no jobs; ready for maintainer review.

